PR #23, opened 2019-11-19
Feature/roslin operator

allanbolipata:
Merge in operator for Roslin

Special appearance commit by `ckandoth`, although that is also me. Go to me if something is broken.


allanbolipata:
Actually need to do #24 still before merging this; gonna try to do it before Monday

runner/operator/tempo_operator/tempo_operator.py
```diff
@@ -13,7 +13,7 @@ def __init__(self, request_id):
         Operator.__init__(self, request_id)
 
     def get_pipeline_id(self):
-        return "1060973e-66e7-45ca-a36d-2b2e26a78d83" # Return ID of the pipeline
+        return "ee5f19d6-f8bf-401f-8cbc-1dd81af10e67" # Return ID of the pipeline
```
sivkovic: I think this would be better to be referenced by github link, entrypoint and version?
Something like:
return `Pipeline.objects.find(github=https://github.com/mskcc/roslin-cwl, version="feature/remove_pairing_file_requirement", entrypoint="workflows/pair-workflow-sv.cwl").first().id()`

allanbolipata: Yeah that's better; let's get these PRs in and definitely let's add that later

PR #77, opened 2020-01-06
Adding singularity definitions

allanbolipata:
... for beagle and celery services.

These should be the set of commits that were lost from a previous PR.


allanbolipata:
May need to update `SINGULARITYENV_` envars to coincide with #76 changes

container/celery_services.def
```diff
@@ -0,0 +1,32 @@
+Bootstrap: docker
+From: ubuntu:19.10
+Includecmd: no
+
+%environment
+    export BEAGLE_PATH=/usr/bin/beagle
+    export CELERY_LOG_PATH=/logs
+    export CELERY_PID_PATH=/tmp
+    export CELERY_BEAT_SCHEDULE_PATH=/tmp
+    export CELERY_EVENT_QUEUE_PREFIX=$(date +%Y%m%dT%H%M%S)
```
allanbolipata: Got this working. Used `CELERY_EVENT_QUEUE_PREFIX` as `-n`; it shows up on `rabbitmq` admin page now with a timestamp by default which can be overwritten.

PR #83, opened 2020-01-08
add Makefile, RabbitMQ conf, fixtures

stevekm:
Add Makefile for setup & management of local Beagle installation. Also add some essential fixtures to populate a fresh Beagle database.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -277,6 +277,25 @@ request:
 	http://$(DJANGO_BEAGLE_IP):$(DJANGO_BEAGLE_PORT)/v0/fs/files/$(REQID)/
 
 # check if the ports needed for services and servers are already in use on this system
+ifeq ($(UNAME), Darwin)
+# On macOS High Sierra, use this command: lsof -nP -i4TCP:$PORT | grep LISTEN
+check-port-collision:
+	@for i in \
+	"DJANGO_BEAGLE_PORT:$(DJANGO_BEAGLE_PORT)" \
+	"BEAGLE_DB_PORT:$(BEAGLE_DB_PORT)" \
+	"RABBITMQ_NODE_PORT:$(RABBITMQ_NODE_PORT)" \
+	"DJANGO_RIDGEBACK_PORT:$(DJANGO_RIDGEBACK_PORT)" \
+	"PGPORT:$(PGPORT)" ; do ( \
+	label="$$(echo $$i | cut -d ':' -f1)" ; \
+	port="$$(echo $$i | cut -d ':' -f2)" ; \
+	lsof -ni | grep LISTEN | tr -s ' ' | cut -d ' ' -f9 | sed -e 's|.*:\([0-9]*\)$$|\1|g' | sort -u | grep -qw "$$port" && echo ">>> $$label port has a collision; something is already running on port $$port" || : ; \
+	) ; done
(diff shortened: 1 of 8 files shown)


Makefile
```diff
@@ -0,0 +1,313 @@
+SHELL:=/bin/bash
```
sivkovic: I would put all of this in for example `dev/` directory, and it would be good that we have README, with explanation how to setup environment and dependencies.

stevekm: if its in a `dev/` dir, it will break many of the file paths and in general make it awkward to use the Makefile and a lot more complicated to configure it. 

The `export` lines show all the environment variables, and the `install` recipe shows the dependency installation, along with the details in the `help` recipe

Makefile
```diff
@@ -0,0 +1,313 @@
...
+
+# remove old RabbitMQ cluster configs;
+# had to do this when RabbitMQ kept crashing due to not finding cluster https://stackoverflow.com/questions/6948624/mnesia-cant-connect-to-another-node
+rabbitmq-clean:
+	rm -rf $(CURDIR)/conda/var/lib/rabbitmq/mnesia/*
+# ./conda/var/lib/rabbitmq/mnesia/rabbit_beagle/cluster_nodes.config
+# ./conda/var/lib/rabbitmq/mnesia/rabbit/cluster_nodes.config
+
+# start all of the Celery worker processes
+celery-start:
```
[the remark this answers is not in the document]
stevekm: I can update the environment variables. Here, the RabbitMQ node name is derived from the current directory basename (`export RABBITMQ_NODENAME:=rabbit_$(CURDIR_BASE)`) so you can have multiple RabbitMQ's for multiple dev instances, given the instance dirname is different, without node name collision.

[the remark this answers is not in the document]
sivkovic: I was just thinking on this command `celery -A beagle_etl worker -l info -Q <beagle_default_queue> -f beagle-worker.log`. Before you didn't need the `-Q` argument. I think in Makefile it still calls this worker without `-Q`.

Makefile
sivkovic: Maybe moving it to the different directory is not the best idea, but I wanted to achieve something like we did with singularity [here](https://github.com/mskcc/beagle/tree/develop/container). Also by README I was thinking on readme similar to one in previous link. Just something which can help user to run beagle using this method as soon as possible without having to read the whole Makefile.

stevekm: I am updating the [`help`](https://github.com/mskcc/beagle/blob/5f793d976fa56ad10c1e866d5ae4cd1470839cea/Makefile#L7) message to make this more clear. `make help` lists the commands needed to set up and run Beagle. Since the Makefile will include all the required environment variables (or as many as possible; `help` will list non-included extras) it will not be required for the user to read the whole thing to get it working. Things like [`check-port-collision`](https://github.com/mskcc/beagle/blob/5f793d976fa56ad10c1e866d5ae4cd1470839cea/Makefile#L282) are also included for the user to quickly troubleshoot issues with services.

Makefile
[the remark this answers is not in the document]
stevekm: changes made in the `test-pairing` branch since its a descendant of this branch

stevekm:
made changes in #91 instead

PR #84, opened 2020-01-09
Script dump request

stevekm:
Added a script to help for development purposes in generation of database fixtures from pre-existing database entries


runner/operator/roslin_operator/bin/test_pair_request.py
```diff
@@ -0,0 +1,211 @@
...
+
+    def test_compile_pairs4(self):
+        """
+        Test pairing with only a single unpaired Tumor sample
+        Test that the appropriate Normal sample is found from the other samples in the same request
+        missing normal for sample 10075_D_1; querying patient C-DRKHP7
+        """
+        # Load fixtures
+        test_files_fixture = os.path.join(settings.TEST_FIXTURE_DIR, "10075_D.file.json")
+        call_command('loaddata', test_files_fixture, verbosity=0)
```
sivkovic: This can't be loaded together with other fixtures same as you did [here](https://github.com/mskcc/beagle/blob/script_dump_request/runner/operator/roslin_operator/bin/test_pair_request.py#L19)? Or there is some other reason you implement it this way?

runner/operator/roslin_operator/bin/test_pair_request.py
```diff
@@ -0,0 +1,211 @@
+import os
```
sivkovic: Also lets create a `runner/tests/` package and put all tests for runner there. It is batter practice to keep all tests at the same place.

And thanks for the tests :)

scripts/db_metrics.py
```diff
@@ -0,0 +1,43 @@
+#!/usr/bin/env python
```
sivkovic: Are we going to need this?

runner/operator/roslin_operator/bin/test_pair_request.py
stevekm: the import [here](https://github.com/mskcc/beagle/blob/script_dump_request/runner/operator/roslin_operator/bin/test_pair_request.py#L19) sets up the base fixtures for all test cases, in order to make the pipeline and file system entries usable. 

This test case is using a specific set of fixtures loaded after that first set to put specific samples' file and filemetadata info into the database.

runner/operator/roslin_operator/bin/test_pair_request.py
stevekm: if you put them all in `runner/tests` will it make it more difficult to track which tests go with which submodules? Especially since we have a heavily nested dir structure here with submodules and functions with the same name across the operators. I usually try to keep tests adjacent to their corresponding `.py` files so its not ambiguous which test file goes with which `.py` file, also makes it a little less complicated to import the module for testing if its adjacent. I think `python manage.py test` automatically discovers all tests?

scripts/db_metrics.py
stevekm: I was using it in order to select requests that had appropriate numbers of items in order to generate fixtures. Not sure where to put it. I am likely going to need it again in the future. Its handy for checking the db metrics quickly.

runner/operator/roslin_operator/bin/test_pair_request.py
sivkovic: You are right `python manage.py test` do discover all the tests. I mean I would prefer following the Django best practice for this, which is keeping all the tests together. In cases you mentioned when you have a nested directory structure, usually what people do is basically mirroring the project structure. In this case I probably wouldn't go all the way to `runner/tests/operator/roslin_operator/bin/test_pair_request.py` but `runner/tests/operator/roslin_operator/test_roslin_operator.py`, where we could put all the testes related to roslin operator. This also helps separating tests by type so you can run only certain tests, which is harder if you have test files in all modules. Something like [this](https://dezoito.github.io/2015/09/21/how-to-test-django-applications_pt1.html)

stevekm:
made changes in other PR

PR #102, opened 2020-01-21
add files ManyToManyField on Port

### LINKED ISSUE #97: Create a Port File table
Create a new Model which will create a connection between File objects and Ports.

stevekm:
Fixes #97


runner/migrations/0013_merge_20200117_1842.py
```diff
@@ -0,0 +1,14 @@
+# Generated by Django 2.2.8 on 2020-01-17 18:42
```
sivkovic: Do we need this migration?

stevekm: it was auto-generated so I assumed its needed to migrate

sivkovic: Yes, I see that. Can you just check do we need it? Because there is no operation in the migration, so if we don't need it I prefer removing it.

PR #104, opened 2020-01-23
Test roslin

stevekm:
added a bunch of needed tests and fixtures to demonstrate functionality


runner/tests/test_port.py
```diff
@@ -0,0 +1,43 @@
...
+    # load fixtures for the test case temp db
+    fixtures = [
+    "file_system.filegroup.json",
+    "file_system.filetype.json",
+    "file_system.storage.json",
+    "runner.pipeline.json"
+    ]
+
+    def test_true(self):
+        self.assertTrue(True)
```
sivkovic: I am not sure I understand this? Why are we testing tests?

runner/tests/test_port.py
```diff
@@ -0,0 +1,43 @@
...
+
+    def test_fixtures1(self):
+        """
```
sivkovic: Same for this one?

stevekm: Trying to make sure that changes to the fixtures json files are recognized as part of the dependency for test. Since the fixtures are saved in external files to the tests, they could be be updated at some point in the future e.g. to reflect model schema changes. This test is trying to make some basic guarantees that the intended number of elements are still saved in the JSON file, otherwise it could affect downstream tests.

runner/tests/test_port.py
stevekm: It's the first test used when setting up a new `TestCase` to help ensure that your new test case is actually getting run. Especially since the current scripted test runner invocation is cherry picking tests to run instead of running them all automatically, it's easy to forget to update the test runner to include the new tests and see a message of "all tests passed" when in fact the new tests were not run. During development, it starts out as `False` to force test failure, then is switched to `True` once you've ensured the test case is running.

runner/tests/test_port.py
sivkovic: Well if fixtures are updated in a way that test no longer works would the test fail? This test looks to me like you have a test to test would another test work. It is a regular thing that you need some kind of a mock data for tests so I think if anybody needs to fix the test which started failing, first thing that person would look at is, fixtures or setUp method. I prefer removing this as well.

runner/tests/test_port.py
sivkovic: My opinion is that this is not needed. You could also delete the test the same way as you switched assert to True. I don't go into how do you do development, but this is just additional test which doesn't give us any information. Considering runner invocation, all tests are fixed on branch feature/fix_locations_in_db_value so you will be able to invoke all the tests really soon. I prefer deleting this.

stevekm:
removed the Port fixtures and tests

PR #133, opened 2020-02-10
Associate pipeline with operator

aef-:


beagle/settings.py
```diff
@@ -278,6 +278,7 @@
 
 POOLED_NORMAL_FILE_GROUP = os.environ.get('BEAGLE_POOLED_NORMAL_FILE_GROUP', '1552617b-3b06-46de-921a-a000a9129385')
 
+CELERY_TASK_ALWAYS_EAGER=DEBUG
```
stevekm: @aef- How does this work? What behavior does it introduce? sounds useful

stevekm:
hey can you try merging the latest develop into this branch and check if the tests work?

beagle/settings.py
aef-: I updated it to use Celery 4 settings but it's exactly what you did in the test files.

aef-:
@stevekm updated so tests passed, thanks for writing them btw, they really helped out.

We may want to look at using pytest and factoryboy in the future when fixtures become a pain to manage due to model changes

lib/admin.py
```diff
@@ -0,0 +1,24 @@
...
+
+def link_relation(field_name):
+    """
+    Converts a foreign key value into clickable links.
+
+    If field_name is 'parent', link text will be str(obj.parent)
+    Link will be admin url for the admin url for obj.parent.id:change
+    """
+    def _link_relation(obj):
+        print(field_name)
```
aef-: will remove.

PR #135, opened 2020-02-11
Outputs metadata

stevekm:
Changes made here in order to pass metadata out of the Roslin operator and set the `output_metadata` field for the job serializer in order to use the output metadata downstream in other operators

### DIFF
--- a/runner/operator/roslin_operator/construct_roslin_pair.py
+++ b/runner/operator/roslin_operator/construct_roslin_pair.py
@@ -5,7 +5,8 @@ from pprint import pprint
 from .bin.make_sample import remove_with_caveats
 from .bin.pair_request import compile_pairs
 
-
+class InvalidAssay(Exception):
+    pass
 
 # TODO: generalize
 def load_references():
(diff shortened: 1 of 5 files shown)


runner/operator/roslin_operator/construct_roslin_pair.py
```diff
@@ -67,7 +67,18 @@ def construct_roslin_jobs(samples):
         job['pair'] = [tumor_sample, normal_sample]
         references = convert_references(project_id, assay)
         job.update(references)
-        roslin_jobs.append(job)
+
+        job_metadata = {}
+        job_metadata['assay'] = assay
+        job_metadata['request_id'] = project_id
+        job_metadata['tumor'] = {}
+        job_metadata['tumor']['igo_id'] = tumor['igo_id']
```
stevekm: @allanbolipata @sivkovic which identifier are we using inside the Roslin pipelines? IGO ID or Patient ID or some other ID?

stevekm:
Once the inclusion of output metadata in the operator is complete, we will have to make sure to check against any unintended side effects in the behavior of the operators and job submission, etc.. Need some integration testing.

stevekm:
note: @nikhil is planning updates to Roslin CWL output schema, changes might need to be included here

Issue #139, opened 2020-02-15
Need method to generate to database fixtures for development


stevekm:
It would help speed up development if we had a method for creating sample datasets for use as example database fixtures more easily. The current method requires loading a backup of an existing database and then dumping fixtures. However this makes it difficult to obtain fixtures that have database fields and values that have been updated in the most recent versions of Beagle but are not yet updated in production database backups. Also, holding production fixtures in your development instance has the side-effect of triggering your dev instance to attempt to resolve the status of prod jobs, and run other automated tasks on the prod data which might be unwanted. 

Since some of the fixtures required to simulate things like pipeline run outputs are very large (a single Roslin pipeline run on one tumor-normal pair currently has more than 250 individual items associated), it becomes difficult to simulate these in a manner needed for further development

related to https://github.com/mskcc/beagle/issues/131

stevekm:
this article suggests some libraries and methods that might be useful for this: https://engineering.giphy.com/fixing-django-fixtures/

Issue #122, opened 2020-01-31
maintain association between Run, Port, File, and FileMetadata entries


stevekm:
The current implementation of FileMetadata uses modification time and `version` to determine which FileMetadata entry to use for a specific file when the file's metadata needs to be accessed, with the highest `version` or most recent timestamp being used. 

This would cause the Run and Port instances from a pipeline run to become disassociated with the FileMetadata for all File's in a pipeline run, since the metadata could be updated later and replaced with a newer version that does not reflect the FileMetadata that was used at the time of the pipeline run. Additionally, if a pipeline run updates a FileMetadata entry with extra fields needed for downstream usage, those fields could be lost if a newer FileMetadata entry is created for a File and does not include them. 

We could avoid this by keeping all four entries associated in a single table, which would hold the foreign keys for each Run, Port, File, and FileMetadata used. This would also reduce the programming logic and number of database queries needed to query things like all File's with certain FileMetadata attributes for a given Run.

stevekm:
another idea for this is to just swap `Port.files` for `Port.filemetadatas`

PR #140, opened 2020-02-15
add fixture duplication script

stevekm:
Created a script to help in producing new unique db fixtures based on old pre-existing ones. 

Hopefully this will help with https://github.com/mskcc/beagle/issues/139

### DIFF
--- a/scripts/duplicate_fixtures.py
+++ b/scripts/duplicate_fixtures.py
@@ -24,6 +24,7 @@ import uuid
 import copy
 import re
 import time
+import argparse
 
 def get_unique_id():
     """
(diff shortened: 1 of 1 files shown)


stevekm:
A note about this script: it is not possible to change the primary key on `File` entries, because there is a unique constraint in the `File` model on [`File.path`](https://github.com/mskcc/beagle/blob/beb81f288d2e933b2a4ebbb8e87df6e7daefd52a/file_system/models.py#L55). 

This results in the problem described in #122. When loading a an old saved `Run` fixture set along with its new duplicated `Run` fixture set created with this script, the two Run's will have the exact same `File` entries due to the unique constraints in the database. Then, when trying to resolve the `FileMetadata` associated with each `Port` item in both the old and new fixture sets, only one `FileMetadata` will be returned because the current method for looking up `FileMetadata` for a given `File` requires returning only the `Filemetadata` with the highest version/most recent date. 

Thus, despite having two unique `Run`'s and sets of `Port`'s and `FileMetadata`'s, it is not possible to actually retrieve the `FileMetadata` for the newly created `Port` items due to the lack of direct association between `Port` and `FileMetadata`.

PR #146, opened 2020-02-19
Operator chaining + admin syntax highlighting

aef-:
I can split up admin syntax highlighting if that makes reviewing this easier.

Tests to be done.


runner/tasks.py
```diff
@@ -14,55 +14,88 @@
...
-    jobs = operator.get_jobs()
-    for job in jobs:
-        if job[0].is_valid():
-            logger.info("Creating Run object")
-            run = job[0].save()
-            logger.info("Run object created with id: %s" % str(run.id))
-            create_run_task.delay(str(run.id), job[1], None)
-        else:
-            logger.error("Job invalid: %s" % str(job[0].errors))
+def process_triggers():
```
aef-: as discussed, this shouldn't be a celery beat, but rather, a single operator run should trigger its dependents when its status is changed to complete. I'm not sure where we should update its status to complete since there are some race conditions at play, so I left this here for now.

runner/tasks.py
```diff
@@ -14,55 +14,88 @@
...
+            logger.info("Trigger %s Fail" % run_id)
+            operator_run.fail()
 
 
 @shared_task
 def create_run_task(run_id, inputs, output_directory=None):
     logger.info("Creating and validating Run")
     try:
         run = RunObject.from_cwl_definition(run_id, inputs)
-        run.ready()
```
aef-: I'll need to add this back.

runner/tasks.py
```diff
@@ -14,37 +14,88 @@
...
+    Run.objects.bulk_update(valid_jobs, operator_run_id=operator_run.id)
+
+    for job in valid_jobs:
+        logger.info("Creating Run object")
+        run = job[0].save()
+        logger.info("Run object created with id: %s" % str(run.id))
+        create_run_task.delay(str(run.id), job[1], None)
+
+    for job in invalid_jobs:
+        logger.error("Job invalid: %s" % str(job[0].errors))
```
aef-: we don't associate invalid runs with an operator run since it would break the conditions -- let me know if this is incorrect.

runner/tasks.py
```diff
@@ -120,6 +171,7 @@ def check_jobs_status():
             remote_status = check_status_on_ridgeback(run.execution_id)
             if remote_status:
                 if remote_status['status'] == 'FAILED':
+                    run.run_operator.update(num_failed_runs)
```
aef-: to remove.

stevekm:
is there an example test case to show how to use the new features?

aef-:
Will add tests shortly.

runner/tasks.py
```diff
@@ -76,6 +129,7 @@ def submit_job(run_id, output_directory=None):
     response = requests.post(settings.RIDGEBACK_URL + '/v0/jobs/', json=job)
     if response.status_code == 201:
         run.execution_id = response.json()['id']
+        run.ready()
```
sivkovic: Why do we have this line here?

runner/tasks.py
sivkovic: Sure.

runner/tasks.py
sivkovic: We might want to report invalid runs somehow. Currently we are just ignoring them. Maybe the error table which we mentioned makes sense for this?

runner/models.py
```diff
@@ -48,6 +95,27 @@ class Run(BaseModel):
...
+            "status": self.status
+        }
+
+    def save(self, *args, **kwargs):
+        # TODO do we want to decrement if a job goes from completed/failed to open or failed to complete?
+        # We can also a prevent a job from going to open once it's in a closed state
+        if self.operator_run and self.original["status"] != self.status:
+            if self.status == RunStatus.COMPLETED:
+                self.operator_run.increment_completed_run()
+                self.original["status"] = RunStatus.COMPLETED
```
sivkovic: I am not sure what is `self.original`?

aef-: It's for this condition here:
https://github.com/mskcc/beagle/pull/146/files/c027c32edba314f9fa800e4090870deb92ade977#diff-e27fa3dbcada827773bbfa1157890a70R110

Let's say we update the status of a run:
```
run.status = RunStatus.COMPLETED
```

There's no way, afaik, to check whether the `status` property was changed.

runner/tasks.py
aef-: That's something we can do, or we can explore something like `logz.io` or even Sentry.

runner/tasks.py
aef-: fixed

Issue #150, opened 2020-02-24
need method for outputting pipeline CWL input format


stevekm:
Right now, the datastructure that pipeline Operators generate from the `get_jobs` method is not the same as the CWL input for the pipeline. In particular, "file" entries are represented like this in the Operator's datastructure:

```
{'class': 'File', 'location': 'bid:///c26e1cf4-0455-4ad7-b2b1-8996cb60214d'}
```

or like this:

```
{'class': 'File', 'location': 'juno:///path/to/foo.txt'}
```

However, the actual CWL representation looks like this:

```
{'class': 'File', 'path': '/path/to/foo.txt'}
```

Currently, Beagle is automatically converting the former two syntaxes into the latter one after the job is submitted (? @sivkovic  please feel free to correct or clarify this). Also, if the user is manually creating the input JSON to pass to the Beagle API to run a pipeline, it must be in the former two syntaxes, and not the latter syntax.

This makes development and manual run creation difficult since the datastructure that you pass to the CWL no longer matches the format used by Beagle, so you cannot verify that your data is correct by running it against the CWL directly. Similarly, when building a new operator, you cannot manually check that your output JSON format is correct since it will not be accepted by `cwltool`, etc., due to the format discrepancies.

It would be helpful to be able to somehow output the exact CWL input JSON format that an operator would be creating in order to validate against the CWL pipeline input format.

PR #151, opened 2020-02-24
Roslin qc

stevekm:
Roslin QC Operator


stevekm:
Fails with `ModuleNotFoundError: No module named 'gitdb.utils.compat'`; @sivkovic what was the fix for this?

sivkovic:
@stevekm fix for this was version bump of GitPython https://github.com/mskcc/beagle/blob/develop/requirements.txt#L11. Fix is already on develop. Did you merge develop back to your branch?

stevekm:
needs to be updated with the new Roslin CWL output format for operator input

stevekm:
This is blocked by https://github.com/mskcc/beagle/issues/154 until we can figure out a way to get copies of the updated Roslin pipeline output format into Beagle.

Harmon758:
> Travis CI Fails with `ModuleNotFoundError: No module named 'gitdb.utils.compat'`; @sivkovic what was the fix for this?

> @stevekm fix for this was version bump of GitPython https://github.com/mskcc/beagle/blob/develop/requirements.txt#L11. Fix is already on develop. Did you merge develop back to your branch?

See https://github.com/gitpython-developers/GitPython/issues/983.

runner/operator/roslin_qc_operator/bin/input.py
```diff
@@ -0,0 +1,395 @@
...
+    """
+    references = json.load(open(references_json))
+    db_files = {}
+    db_files['fp_genotypes'] = references["targets"][assay]['FP_genotypes']
+    db_files['hotspot_list_maf'] = references["request_files"]['hotspot_list_maf']
+    db_files['conpair_markers'] = references["request_files"]['conpair_markers']
+    db_files['ref_fasta'] = references["request_files"]['ref_fasta']
+
+    # need to resolve some items to URI's
+    for key in ['fp_genotypes', 'hotspot_list_maf', 'ref_fasta']:
```
sivkovic: Are these files specified in references_json `runner/operator/roslin_operator/reference_jsons/roslin_resources.json`? If they are why do we have this logic to check is it a path or uri?

runner/operator/roslin_qc_operator/roslin_qc_operator.py
```diff
@@ -0,0 +1,81 @@
+from runner.operator.operator import Operator
+from runner.models import Run, Pipeline, Run,  RunStatus
+from runner.serializers import APIRunCreateSerializer
+from .bin import input
+import datetime
+
+class InvalidPipelineError(Exception):
```
sivkovic: I don't think we need this.

runner/operator/roslin_qc_operator/roslin_qc_operator.py
```diff
@@ -0,0 +1,81 @@
...
+        self.runs = None
+        if self.run_ids != None:
+            self.runs = Run.objects.filter(id__in = self.run_ids)
+        # TODO: add support for things like FileMetadata's
+        # TODO: add support for multiple request ids
+
+        Operator.__init__(self, model, request_id = request_id, run_ids = run_ids)
+
+    def get_pipeline_id(self):
+        return "9b7f2ac8-03a5-4c44-ae87-1d9f6500d19a"
```
sivkovic: I don't think we need this anymore as well.

runner/tasks.py
```diff
@@ -95,6 +96,9 @@ def process_triggers():
 @shared_task
 def create_run_task(run_id, inputs, output_directory=None):
     logger.info("Creating and validating Run")
+    if settings.DUMP_JSON:
```
sivkovic: I don't like putting this in a task. Prints like this for debug are fine, but I think we should remove them from production code. I don't like the idea of generating files in which are not used anywere.

runner/views/run_api_view.py
```diff
@@ -92,13 +93,25 @@ class OperatorViewSet(GenericAPIView):
...
-        request_ids = request.data['request_ids']
+        request_ids = request.data.get('request_ids', None)
+        run_ids = request.data.get('run_ids', None)
         pipeline_name = request.data['pipeline_name']
         pipeline = get_object_or_404(Pipeline, name=pipeline_name)
 
-        for request_id in request_ids:
-            logging.info("Submitting requestId %s to pipeline %s" % (request_id, pipeline_name))
-            create_jobs_from_request.delay(request_id, pipeline.operator_id)
+        if pipeline_name == "roslin-qc":
```
sivkovic: Why aren't we handling roslin-qc operator the same way as any other operator? request_handlers shouldn't have any operator specific logic in them.

sivkovic (changes requested):

runner/operator/roslin_qc_operator/roslin_qc_operator.py
stevekm: I tried to remove it and things started breaking. Not sure why I think it has to do with the recent changes to how Operators work.

runner/tasks.py
stevekm: This is required for development, debug, and manual run setup as per https://github.com/mskcc/beagle/issues/162 and https://github.com/mskcc/beagle/issues/150. Otherwise there is no straightforward method to actually see what your Job JSON and final output CWL look like. Its easier to leave it in as a feature than to have to re-add it every single time you need to use it for debugging.

stevekm: This is required for development, debug, and manual run setup as per https://github.com/mskcc/beagle/issues/162 and https://github.com/mskcc/beagle/issues/150. Otherwise there is no straightforward method to actually see what your Job JSON and final output CWL look like. Its easier to leave it in as a feature than to have to re-add it every single time you need to use it for debugging. There are probably better ways to implement this but it will take a lot more time to refactor and build.

runner/views/run_api_view.py
stevekm: because Roslin QC Operator requires that all `request_ids` and `run_ids` be passed, while other Operators only currently support a single `request_id` arg. Trying to refactor all other Operators to be able to use this would take a lot more work. Also, its less confusing when the logic for dictating the behavior of how the handling takes place is placed up front here near the endpoint/view, instead of buried way down inside some other module.

runner/operator/roslin_qc_operator/bin/input.py
```diff
@@ -0,0 +1,395 @@
...
+    """
+    # get the base datastructure dict for each R1 and R2
+    normal_r1 = build_sample(pair['normal']['R1'])
+    normal_r2 = build_sample(pair['normal']['R2'])
+    tumor_r1 = build_sample(pair['tumor']['R1'])
+    tumor_r2 = build_sample(pair['tumor']['R2'])
+
+    # merge the dict's; the second dict will overwrite args from the first
+    normal = {**normal_r1, **normal_r2}
+    normal['R1'] = file_to_cwl(pair['normal']['R1'].file)
```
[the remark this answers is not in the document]
stevekm: This was not clear when I first started building the Operator, but the code works well for producing the CWL output needed in order to develop the Operator so I am not sure what to do with it now that the Operator has been refactored to output Job JSON instead of CWL JSON

runner/tests/serializers/test_serializers.py
```diff
@@ -0,0 +1,64 @@
...
+    "runner.pipeline.json",
+    "beagle_etl.operator.json"
+    ]
+
+    def test_create_run_serializer1(self):
+        """
+        Test that the API Run Create Serializer works and creates a Run
+        """
+        # start with 0 runs in the database
+        self.assertEqual(len(Run.objects.all()), 0)
```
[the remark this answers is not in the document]
stevekm: if you POST to `<localhost:1234>/v0/run/api/` you are going to get the real server, not the test-instance.

stevekm:
@sivkovic the copy outputs operator is failing the tests, I am not sure what is wrong with it; https://travis-ci.com/mskcc/beagle/builds/151844563#L2014

PR #156, opened 2020-02-24
remove task_always_eager

sivkovic:


beagle_etl/celery.py
```diff
@@ -17,7 +17,7 @@
 # Load task modules from all registered Django app configs.
 app.autodiscover_tasks()
 
-app.conf.task_always_eager = settings.DEBUG
+# app.conf.task_always_eager = settings.DEBUG
```
stevekm: was this the setting that allowed for Celery tasks to run synchronously during debugging? That was really helpful for dev and debug

PR #158, opened 2020-02-25
Added check for None type in pooled normals resolution

allanbolipata:
I was still getting errors on my local box, even with the change [here](https://github.com/mskcc/beagle/blob/develop/beagle_etl/jobs/lims_etl_jobs.py#L105).

This was happening when I did a couple import requests, like `10058_B`.


allanbolipata:
This was the error 
```
File "/home/bolipatc/dev_voyager/beagle/beagle_etl/tasks.py", line 72, in process
    self._process()
   File "/home/bolipatc/dev_voyager/beagle/beagle_etl/tasks.py", line 97, in _process
    children = func(**self.job.args)
   File "/home/bolipatc/dev_voyager/beagle/beagle_etl/jobs/lims_etl_jobs.py", line 107, in fetch_samples
    for f in pooled_normals:
```

PR #173, opened 2020-03-04
Operator runs fix

sivkovic:
When all runs from OperatorRun completes if there is no specified triggers it will stay in RUNNING state forever. Adding check if there is no triggers, to set the OperatorRun instance in proper state. Not sure is this the best way, but it can do it's purpose for now.

### DIFF
--- a/runner/tasks.py
+++ b/runner/tasks.py
@@ -67,7 +67,6 @@ def process_triggers():
                     condition = trigger.aggregate_condition
                     if condition == TriggerAggregateConditionType.ALL_RUNS_SUCCEEDED:
                         if operator_run.percent_runs_succeeded == 100.0:
-                            operator_run.complete()
                             create_jobs_from_chaining.delay(
                                 trigger.to_operator_id,
                                 trigger.from_operator_id,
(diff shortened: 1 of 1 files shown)


runner/tasks.py
```diff
@@ -92,6 +92,12 @@ def process_triggers():
                     if operator_run.percent_runs_finished == 100.0:
                         operator_run.complete()
 
+            if operator_run.percent_runs_finished == 100.0:
```
aef-: I think this should go in a conditional checking whether there are any triggers.

aef-:
Good catch, I think we need to have a one-to-many relationship between OperatorRun and OperatorRunStatus for each Trigger.

90% may pass when 100% may fail. What should appear? How is this shown to the admin in beagle?

runner/tasks.py
sivkovic: Yes. That is why I added it. My second OperatorRun was hanging in RUNNING state.

aef-: I meant a conditional to check if the operator run has any triggers associated with it, but actually i think we should keep this and remove all other `operator_run.complete/fail()`

sivkovic: @aef- Something like this?

PR #198, opened 2020-03-25
Feature/expand pairing logging and rg id fix

allanbolipata:


allanbolipata:
Merged even without approval because https://github.com/mskcc/beagle/pull/205 had its commits and was approved.

PR #212, opened 2020-04-02
Feature/qc co output dir

allanbolipata:
This makes the QC and Copy Outputs operators write to the same directory, provided that "project prefix" (or request Id) exists.

If job group ID also exists, it appends that to the end as another subdirectory.


runner/serializers.py
```diff
@@ -141,12 +141,21 @@ def create(self, validated_data):
...
-        run = Run(name=name, app=pipeline, status=RunStatus.CREATING, job_statuses=dict(), output_metadata=validated_data.get('output_metadata', {}), tags=tags)
+        if validated_data.get('output_directory') is not None:
+            output_directory = validated_data.get('output_directory')
+        run = Run(name=name,
+                app=pipeline,
+                status=RunStatus.CREATING,
+                job_statuses=dict(),
+                output_metadata=validated_data.get('output_metadata', {}),
+                tags=tags,
+                output_directory = output_directory)
```
sivkovic, line 158: You could just put here `validated_data.get('output_directory')` and avoid having https://github.com/mskcc/beagle/pull/212/files#diff-d09e3a55b7e8db17d6d3d12926580e74R150. But it is fine either way.

PR #251, opened 2020-05-05
Access FASTQ to Bam Operator

aef-:


allanbolipata:
@aef- I added another operator and that introduced a conflict with this one.

We are making another release next week - if after the conflict is resolved and you are not making any more changes to this ticket beyond that, we can merge it before then

aef-:
Sounds good

PR #256, opened 2020-05-07
Add recipe filter to jobs admin page

aef-:
I find myself searching for access jobs a lot on production, this would make that much easier.

Screen Shot 2020-05-07 at 15 07 32

### DIFF
--- a/beagle_etl/admin.py
+++ b/beagle_etl/admin.py
@@ -1,4 +1,6 @@
+import os
 from django.contrib import admin
+from django.contrib.admin import ModelAdmin, SimpleListFilter
 from django.utils.safestring import mark_safe
 from .models import Job, JobStatus, Operator
 from lib.admin import pretty_python_exception, pretty_json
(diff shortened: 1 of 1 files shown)


beagle_etl/admin.py
```diff
@@ -31,8 +50,18 @@ class JobAdmin(admin.ModelAdmin):
...
+    list_filter = (RecipeFilter, )
+
+    def get_short_run(self, obj):
+        if obj.run:
+            (_, run) = os.path.splitext(obj.run)
+            return mark_safe("<span title='%s'>%s</span>" % (obj.run, run[1:]))
+        else:
+            return '--'
+
+    get_short_run.short_description = 'Run'
```
aef-, line 62: this shortens the run name, you can hover over the text for the full path.

beagle_etl/admin.py
```diff
@@ -18,9 +20,26 @@ def restart(modeladmin, request, queryset):
...
 
-class JobAdmin(admin.ModelAdmin):
-    list_display = ('id', 'run', 'retry_count',
+    def lookups(self, request, model_admin):
+        options = set()
+
+        for o in Operator.objects.values("recipes"):
+            for recipe in o["recipes"]:
+                options.add((recipe, recipe))
+        return options
```
aef-, line 33: a caveat is that it doesn't catch all recipes coming from LIMS, just those used in Operators

PR #270, opened 2020-05-12
Operator/helix filters 01

allanbolipata:


runner/operator/helix_filters/construct_helix_filters_input.py
```diff
@@ -0,0 +1,161 @@
...
+
+
+def get_file_obj(file_obj):
+    """
+    Given file_obj, construct a dictionary of class File, that file's
+    JUNO-specific URI file path, and a list of secondary files with
+    JUNO-specific URI file paths
+    """
+    secondary_file_list = []
+    file_location = file_obj['location'].replace('file://', '')
```
sivkovic, line 65: I think I mentioned this in some previous PR we should try to use `FileProcessor` for file manipulation. Most of the methods for extracting file paths from location etc is already implemented there.

allanbolipata: Ah, right. I copied this over from some other CWL

PR #279, opened 2020-05-13
Seqosystem notifier

aef-:
Sends ACCESS workflow updates to SEQOSYSTEM

Work still to be done.


allanbolipata:
@aef- We will move this to `1.7.0` milestone since it won't make today's release

PR #283, opened 2020-05-13
Feature/run api list refactor

nikhil:
- Better integration with Swagger
- Added additional query parameters
- Allow request ids to be a list
- Changed inputs query to ports
- Implement schema validation through serializers
- Added tests
- Resolves: #272


runner/views/run_api_view.py
```diff
@@ -24,41 +26,91 @@ class RunApiViewSet(mixins.ListModelMixin,
...
+        serializer = RunApiListSerializer(data=fixed_query_params)
+        if serializer.is_valid():
+            queryset = time_filter(Run, fixed_query_params)
+            job_groups = fixed_query_params.get('job_groups')
+            jira_ids = fixed_query_params.get('jira_ids')
+            status_param = fixed_query_params.get('status')
+            ports = fixed_query_params.get('ports')
+            tags = fixed_query_params.get('tags')
+            request_ids = fixed_query_params.get('request_ids')
+            created_before = fixed_query_params.get('created_before')
```
sivkovic, line 67: What is the difference between this and `time_filter`? Also we don't have query params required in `time_filter` listed in RunApiListSerializer.

nikhil: Good catch, I missed that function. It should be fixed now.

PR #296, opened 2020-05-16
Feature/fix lims import

allanbolipata:
This should import all fastqs now

Has commits from #294, I used it for debugging.


beagle_etl/jobs/lims_etl_jobs.py
```diff
@@ -331,39 +331,31 @@ def fetch_sample_metadata(sample_id, igocomplete, request_id, request_metadata):
...
-                file_search = FileRepository.filter(path=fastqs[1]).first()
-                if not file_search:
-                    create_file(fastqs[1], request_id, settings.IMPORT_FILE_GROUP, 'fastq', igocomplete, data, library, run,
-                                request_metadata, R1_or_R2(fastqs[1]))
-                else:
-                    logger.error(
-                        "File %s already created with id:%s" % (file_search.file.path, str(file_search.file.id)))
-                    conflict = True
-                    conflict_files.append((file_search.file.path, str(file_search.file.id)))
-    if conflict:
```
allanbolipata, line 359: Removed this piece so that the runs don't get reported as a FAILED if we see the file again during the same import but we may need to think about how to handle this better

PR #304, opened 2020-05-18
Roslin rename

nikhil:
Giant PR to incorporate migration from roslin to argos


allanbolipata:
Test.

After this, we need to register new pipelines in staging and test that

nikhil:
Resolves: #278

Issue #199, opened 2020-03-25
Reduce types of recipes for pooled normal import to a controlled list


allanbolipata:
We've been doing string parsing to assign `baitSet`/`recipe` values for pooled normals from the file names, a field necessary for pairing. The parsing, however, has been having issues when the file names are split by an unexpected delimiter.

For example, `HemePACT_v4` would be registered as a `HemePACT` recipe because of the underscore (the file name has multiple underscores). This causes issues when pairing as we're searching for `HemePACT_v4`, the value that's in the sample record.

Let's reduce the data to a controlled list of recipes/bait sets that we can add to in the future.

allanbolipata:
Recipe | Number of Pooled Normals Found with Recipe String
---- | ----
AmpliSeq | 0
ddPCR | 0
RNASeq | 0
RNASeq_PolyA | 0
BRAINPACT_v1 | 12
Archer-HemePanel | 0
WholeGenomeSequencing | 0
HemePACT_v3 | 260
CustomAmplificationPCR | 0
Archer-SolidTumorPanel | 0
M-IMPACT | 164
96Well_SmartSeq2 | 0
WholeGenomeBisulfateSequencing | 0
IMPACT468 | 848
IWG | 0
HumanWholeGenome | 0
IDT_Exome_v1_FP | 0
(Multiple of the Above) | 0
MethylCaptureSeq | 20
PanCancerV2 | 0
M-IMPACT_v1 | 164
MouseWholeGenome | 0
08050 | 4
Archer-Immunoverse | 0
Spo11Oligo | 0
HemeBrainPACT | 12
AMBIGUOUS | 16
AmpliconSeq | 0
CRISPRSceen | 0
MSK-ACCESS_v1 | 0
10X_Genomics_GeneExpression | 0
IonAmpliseqCancerHotspotv2 | 0
IMPACT410 | 654
ShallowWGS | 0
InDropSeq | 0
HemeBrainPACT_v1 | 12
RRBS | 0
SMARTerAmpSeq | 0
MissionBio-Heme | 0
RNASeq_RiboDeplete | 0
Agilent_v4_51MB_Human | 0
HemePACT | 520
07220_Hg19 | 0
CRISPRScreen | 0
ATACSeq | 0
CustomCapture | 12
SingleCellCNV | 0
Agilent_MouseAllExonV1 | 0
WholeExomeSequencing | 4
CRISPRSeq | 0
06605_F_Hg19 | 0
HemePACT_v4 | 246
[...]

PR #329, opened 2020-05-28
Feature/pooled normal recipe list

nikhil:
Resolves: #199


beagle_etl/jobs/lims_etl_jobs.py
```diff
@@ -244,22 +244,28 @@ def create_pooled_normal(filepath, file_group_id):
         logger.info("Pooled normal already created filepath")
     file_group_obj = FileGroup.objects.get(id=file_group_id)
     file_type_obj = FileType.objects.filter(name='fastq').first()
+    assays = Assay.objects.first()
```
allanbolipata, line 247: Just commenting here because I have a feeling we will change this in the future

PR #361, opened 2020-06-08
Added assay view

nikhil:


beagle_etl/views.py
```diff
@@ -77,6 +77,14 @@ def fix_query_list(self, request_query, key_list):
                     query_dict[single_param] = query_value
         return query_dict
 
+class AssayViewSet(mixins.CreateModelMixin,
+                 mixins.DestroyModelMixin,
```
sivkovic, line 81: Assay is basically model which is used as configuration, and we shouldn't have Create and Destroy endpoints. We only need to get values from that record, and update it, we might implement this as a GET, and POST `/v0/assay` where we will return values on GET, and update them on POST. We should also add validation on POST, that all assays specified in `disabled` and `hold`, needs to be listed in `all` as well, and same assay shouldn't be in both `disabled` and `hold`.

beagle_etl/views.py
```diff
@@ -77,6 +77,49 @@ def fix_query_list(self, request_query, key_list):
...
+        return Response(error_message,status=status.HTTP_404_NOT_FOUND)
+
+    @swagger_auto_schema(request_body=AssayUpdateSerializer,responses={200: AssayElementSerializer})
+    def post(self, request):
+        all_list = request.data.getlist('all')
+        disabled_list = request.data.getlist('disabled')
+        hold_list = request.data.getlist('hold')
+        assay_list = Assay.objects.all()
+        if assay_list and len(assay_list) == 1:
+            assay = assay_list[0]
```
sivkovic, line 101: This could be done as `Assay.objects.first()`. It will return only first element.

beagle_etl/views.py
```diff
@@ -77,6 +77,49 @@ def fix_query_list(self, request_query, key_list):
...
+            if all_list:
+                assay.all = list(set(all_list))
+            if disabled_list:
+                assay.disabled = list(set(disabled_list))
+            if hold_list:
+                assay.hold = list(set(hold_list))
+            for single_assay in assay.hold:
+                if single_assay in assay.disabled:
+                    error_message = "Assay {} is in both disabled and hold".format(single_assay)
```
sivkovic, line 110: Only thing which would be really good is going through all assays and join them so error message have list of all assays which are in disabled and hold and also if they missing in the list of all assays. This would enable us to fix all errors if there are more then one.

beagle_etl/views.py
nikhil: I was using ```len(assay_list) == 1``` to fail if for some reason the assay object list is greater than 1

beagle_etl/views.py
nikhil: Yeah, that makes sense. I will make it return a list of errors.

Issue #357, opened 2020-06-06
Admin Hold state in JIRA should notify/CC admins


allanbolipata:
http://jira.mskcc.org:8090/projects/VA/issues/VA-135?filter=allopenissues
http://jira.mskcc.org:8090/projects/VA/issues/VA-118?filter=allopenissues
http://jira.mskcc.org:8090/projects/VA/issues/VA-136?filter=allopenissues

It moves to Admin Hold but it is unclear if it notifies anyone

PR #381, opened 2020-06-15
CC admins when in AdminHold state

sivkovic:
## Changelog

- #357 CC admins in case ticket goes to AdminHold

### DIFF
--- a/notifier/events/custom_capture_cc_event.py
+++ b/notifier/events/custom_capture_cc_event.py
@@ -1,0 +1,23 @@
+from django.conf import settings
+from notifier.event_handler.event import Event
+
+
+class CustomCaptureCCEvent(Event):
+
+    def __init__(self, job_group, assay):
+        self.job_group = job_group
+        self.assay = assay
+
+    @classmethod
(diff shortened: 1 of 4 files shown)


notifier/events/custom_capture_cc_event.py
```diff
@@ -0,0 +1,23 @@
...
+    def get_type(cls):
+        return "CustomCaptureCCEvent"
+
+    @classmethod
+    def get_method(cls):
+        return "process_custom_capture_cc_event"
+
+    def __str__(self):
+        TEMPLATE = """
+        Ticket in AdminHold because of the assay: {assay}. {cc}
```
nikhil, line 21: I think maybe it should be CustomCaptureCCEvent instead of AdminHold event

nikhil: The JIRA states are fine

sivkovic: It can be more assays then CustomCapture that can put ticket to admin hold. But we will put assay in comment as well.

Issue #393, opened 2020-06-17
QCPDF uploaded to notifier (JIRA) is corrupt


allanbolipata:
PDF outputted by ARGOS is correct on Juno file system but the file uploaded to JIRA is corrupted.

PR #408, opened 2020-06-22
JIRA attachment production problem

sivkovic:
## Changelog

- #393 Problem with uploading QC was because `/juno/work/ci` wasn't mounted. In order to standardize the way we are running the staging and production we should use `bin/run_beagle.sh` script.


bin/run_beagle.sh
```diff
@@ -0,0 +1,34 @@
...
+
+echo "Stopping $1 environment"
+singularity instance stop $1_beagle_services
+singularity instance stop $1_beagle_celery
+
+echo Starting $1 environment version $3
+
+if [ $1 != "production" ]; then
+    singularity instance start --containall --bind /srv/services/voyager/ --bind /home/ivkovic/rabix-cli-1.0.5 --bind /tmp --bind /ifs --bind /juno/work/ci /srv/services/voyager/beagle_service_%3.sif $1_beagle_services
+    singularity instance start --containall --bind /srv/services/voyager/ --bind /home/ivkovic/rabix-cli-1.0.5 --bind /tmp --bind /ifs --bind /juno/work/ci /srv/services/voyager/beagle_celery_%3.sif $1_beagle_celery
```
allanbolipata, line 30: `%3`? I think change that to `$3`

sivkovic: Oh thank you.

Issue #372, opened 2020-06-15
Add Job Group id value and put it in notifier/JIRA ticket


allanbolipata:
This will help faster fetching of info.

Issue #371, opened 2020-06-15
Runs api doesn't display full status


allanbolipata:
Right now it just displays the tags information

Issue #370, opened 2020-06-15
Add notifier for FAILED runs


allanbolipata:
CREATED and COMPLETED runs get a comment on the JIRA notifier; need one for FAILED, too.

Issue #367, opened 2020-06-11
Add migration to update finished_date on previous jobs


nikhil:
Lets set the finished_date to last_modified date if it does not exist

PR #413, opened 2020-06-24
Release/1.8.0

allanbolipata:
Changelog
--
**Enhancements**
Add notifier for FAILED runs #370 #401 
Add Job Group id value and put it in notifier/JIRA ticket #372 
Added Job Group Id to jira ticket #409
Add migration to update finished_date on previous jobs #367
When a request contains only normal samples, send to Can't Do state and apply all_normals label #362 
Retrieve error messages from Ridgeback #241 

**Bug fixes**
Runs api doesn't display full status bug #371
Feature/fix run api view #406
Wrong log path for a failed job #231 
Fix operator start runs #410 
Feature/fie view fixes and distribution view #407


runner/serializers.py
```diff
@@ -140,7 +140,7 @@ def get_status_url(self, obj):
 
     class Meta:
         model = Run
-        fields = '__all__'
+        fields = ('id', 'name', 'status', 'tags', 'app', 'inputs', 'outputs', 'status_url', 'created_date', 'job_statuses','execution_id','output_metadata','output_directory','operator_run','job_group','notify_for_outputs','finished_date')
```
sivkovic, line 143: `message` should be included in RunSerializerFull as well, since we are currently not able to see the message when we do a GET /v0/run/api. I don't think it is worth hotfixing, but lets make the change for 1.9.0.

Issue #387, opened 2020-06-16
Update "update request" etl job


allanbolipata:
We have to discuss what this etl job needs to do since it has been left unattended.

For now, it needs to do the following:
- [x] Needs to be able to handle adding fastqs to previously imported request (assumption: data is re-delivered on IGOLIMS and shows up in the usual endpoint)

allanbolipata:
This happened as recently as June 19: request `10848_B` got new fastqs.

Voyager has two samples from this request, with each sample containing two fastqs. On IGOLIMS, `10848_B_3` itself has six fastqs.

allanbolipata:
We will need to add logic for processing re-deliveries, create new ticket and process them automatically

PR #417, opened 2020-06-26
Implement handling re-delivered samples

sivkovic:
## Changelog

- #387 Handle re-delivered samples
    - If file doesn't exist create a new one
    - If file already exist and metadata is different then it was, update metadata


beagle_etl/jobs/lims_etl_jobs.py
```diff
@@ -193,22 +193,74 @@ def get_or_create_pooled_normal_job(filepath, job_group=None):
...
+
+# def create_redelivery_sample_job(sample_id, igocomplete, request_id, request_metadata, job_group=None):
+#     job = Job.objects.create(run=TYPES['REDELIVERY_SAMPLE'],
+#               args={'sample_id': sample_id, 'igocomplete': igocomplete, 'request_id': request_id,
+#                     'request_metadata': request_metadata, 'redelivery': True},
+#               status=JobStatus.CREATED,
+#               max_retry=1, children=[],
+#               job_group=job_group)
+#     return job
+
```
allanbolipata, line 262: Should we remove these

PR #428, opened 2020-07-01
Feature/update helix filters operator

allanbolipata:
Helix filters operator updated to work with most recent version

Added new resources `known_fusions_filter.txt` and `oncotree.json`, which is used by the oncotree data handler if it can't get an internet connection to query oncotree directly.


notifier/helper.py
```diff
@@ -11,7 +11,9 @@ def get_project_id(request_id):
 def generate_sample_data_content(files, pipeline_name, pipeline_github, pipeline_version):
     result = "SAMPLE_ID\tREQUEST_ID\tPROJECT_ID\tPATIENT_ID\tCOLLAB_ID\tSAMPLE_TYPE\tGENE_PANEL\tONCOTREE_CODE\tSAMPLE_CLASS\tSPECIMEN_PRESERVATION_TYPE\tSEX\tTISSUE_SITE\tIGO_ID\tPIPELINE_NAME\tPIPELINE_GITHUB_LINK\tPIPELINE_VERSION\n"
     ret_str = 'metadata__sampleId'
-    query = Q(file__path__in=files) & Q(file__file_group_id=settings.IMPORT_FILE_GROUP)
+    query = Q(file__file_group_id=settings.IMPORT_FILE_GROUP)
+    query |= Q(file__file_group__slug="origin-unknown")
```
allanbolipata, line 15: This assumes the file, if found, are structured the same way as in the LIMS filegroup.

This is needed because I suspect non-LIMS data sources will go through Helix Filters, requiring data clinical files.

PR #462, opened 2020-07-23
Feature/add cwl ready inputs and outputs

nikhil:
Resolves: #457

### DIFF
--- a/runner/run/processors/port_processor.py
+++ b/runner/run/processors/port_processor.py
@@ -3,6 +3,7 @@ import uuid
 import copy
 import logging
 from enum import Enum
+from pathlib import Path
 from runner.run.processors.file_processor import FileProcessor
 from runner.exceptions import PortProcessorException, FileConflictException, FileHelperException
 from notifier.tasks import send_notification
@@ -16,6 +17,7 @@ class PortAction(Enum):
     CONVERT_TO_PATH = 1
     REGISTER_OUTPUT_FILES = 2
     SEND_AS_NOTIFICATION = 3
+    CONVERT_TO_CWL_FORMAT = 4
     FIX_DB_VALUES = 99 # Temporary for fixing values in DB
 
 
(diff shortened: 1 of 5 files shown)


runner/serializers.py
```diff
@@ -164,6 +174,30 @@ class Meta:
         fields = ('id', 'name', 'status', 'tags', 'app', 'inputs', 'outputs', 'status_url', 'created_date', 'job_statuses','execution_id','output_metadata','output_directory','operator_run','job_group','notify_for_outputs','finished_date','message')
 
 
+class RunSerializerCWLInput(RunSerializerPartial):
+
+    inputs = serializers.SerializerMethodField()
+
+    def get_inputs(self, obj):
+        input_data = PortSerializer(obj.port_set.filter(port_type=PortType.INPUT).all(), many=True).data
+        return format_port_data(input_data)
```
sivkovic, line 183: How should the File object look like? Do we need path/location/size etc, and should it be runnable by Voyager or locally?

nikhil: As previously discussed, we will create a uniform format for inputs and outputs

runner/run/processors/port_processor.py
```diff
@@ -121,6 +125,37 @@ def _convert_to_path(val):
...
+        location = file_obj.pop('location',None)
+        if location:
+            try:
+                file_db_object = FileProcessor.get_file_obj(location)
+            except FileHelperException as e:
+                raise PortProcessorException('File %s not found' % location)
+            path = file_db_object.path
+            if path:
+                path_obj = Path(path)
+                if not file_obj.get('checksum'):
```
sivkovic, line 140: I don't think we should check file_obj for checksum because, when we register output files we are always saving the checksum [here](https://github.com/mskcc/beagle/blob/master/runner/run/processors/port_processor.py#L159). [This](https://github.com/mskcc/beagle/pull/462/files#diff-7391770a3e6b593d59c60681987308cfR141-R143) is good enough. Same thing for input files, we are using values from File object in DB to create a job.

runner/run/processors/port_processor.py
```diff
@@ -121,6 +125,37 @@ def _convert_to_path(val):
...
+    def _covert_to_cwl_format(val):
+        file_obj = copy.deepcopy(val)
+        location = file_obj.pop('location',None)
+        if location:
+            try:
+                file_db_object = FileProcessor.get_file_obj(location)
+            except FileHelperException as e:
+                raise PortProcessorException('File %s not found' % location)
+            path = file_db_object.path
+            if path:
```
sivkovic, line 138: [File instance](https://github.com/mskcc/beagle/blob/master/file_system/models.py#L73) can't exist without path. null=False, and required=True is by default. So this condition will always be satisfied.

PR #470, opened 2020-07-31
removes newline chars from within tissue location string

allanbolipata:
This should get the sample data clinical to be made properly for helix filters, but let's discuss https://github.com/mskcc/LimsRest/issues/137 before we merge this in.

Waiting on guidance from IGODT.

### DIFF
--- a/notifier/helper.py
+++ b/notifier/helper.py
@@ -29,7 +29,7 @@ def generate_sample_data_content(files, pipeline_name, pipeline_github, pipeline
             metadata['specimenType'],
             metadata['preservation'],
             metadata['sex'],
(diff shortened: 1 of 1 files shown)


allanbolipata:
Let's get this in so we can run some outstanding projects.

This is expected to be overridden by #474

nikhil:
I think this might not work since .replace returns the new string and it does not modify in place

nikhil:
```
>>> sample_str = "remove_me"
>>> sample_str.replace("remove_me","")
''
>>> sample_str
'remove_me'

```

nikhil:
oh nvm, I expanded the the code segment and it looks fine

Issue #458, opened 2020-07-23
Inconsistent behavior when using values_metadata query


allanbolipata:
When I query by `C-VX4AA7`, I get the following sample names

```
GET /v0/fs/files/?metadata=patientId:C-VX4AA7&values_metadata=sampleName
HTTP 200 OK
Allow: GET, POST, HEAD, OPTIONS
Content-Type: application/json
Vary: Accept

{
    "next": null,
    "previous": null,
    "count": 4,
    "last": 1,
    "results": [
        "C-VX4AA7-M001-d",
        "C-VX4AA7-M002-d",
        "C-VX4AA7-N001-d",
        "SK_MEL_1023A_CL_P14-09483C"
    ]
```

... and I get the following sample classes

```
GET /v0/fs/files/?metadata=patientId:C-VX4AA7&values_metadata=sampleClass
HTTP 200 OK
Allow: GET, POST, HEAD, OPTIONS
Content-Type: application/json
Vary: Accept

{
    "next": null,
    "previous": null,
    "count": 2,
    "last": 1,
    "results": [
        "Metastasis",
        "Normal"
    ]
```

But when I ask to return both sample name and sample class, I get

```
GET /v0/fs/files/?metadata=patientId:C-VX4AA7&values_metadata=sampleName,sampleClass
HTTP 200 OK
Allow: GET, POST, HEAD, OPTIONS
Content-Type: application/json
Vary: Accept

{
    "next": null,
    "previous": null,
    "count": 4,
    "last": 1,
    "results": [
        [
            "SK_MEL_1023A_CL_P14-09483C",
            "Metastasis"
        ],
        [
            "SK_MEL_1023A_CL_P14-09483C",
            "Metastasis"
        ],
        [
            "C-VX4AA7-M001-d",
            "Metastasis"
        ],
        [
            "C-VX4AA7-M001-d",
            "Metastasis"
        ]
    ]
}
```

I lose a reference to the normal and also and one of the tumor samples. Is this expected behavior?

PR #479, opened 2020-08-12
Fix/distinct query

nikhil:
Resolves: #458


file_system/repository/file_repository.py
```diff
@@ -86,6 +86,6 @@ def filter(cls, queryset=None, path=None, path_in=[], path_regex=None, file_type
         if values_metadata_list:
             values_metadata_query_list = ['metadata__%s' % single_metadata for single_metadata in values_metadata_list ]
```
sivkovic, line 87: Maybe we can add order_by query parameter as a future improvement so we can have the option to choose how to order elements. But for now this is good!

PR #563, opened 2020-10-16
Ij/access snv operator

ionox0:
SNV operator for ACCESS V1 `snps_and_indels.cwl` workflow

Todo:
- [x] Django / python unittest wrapper for fixtures import / operator test
- [x] Create new FileGroup for access curated Bams
- [x] Remove hard-coded paths to reference bams (import them into Beagle using API and query for them in the operator)
- [x] how to filter FileRepo for Bam files only?
- [x] use the most recently-sequenced matched normal
- [x] access_curated_bams FileGroup should be queried by slug instead of ID (so it works across environments)
- [x] Look for past tumor samples that have not been run through SNV pipeline due to missing normal and run them when it becomes available
- [x] Include a run tag with sample IDs so that we can filter for samples that haven't been run when this operator runs (solves this issue ^)
- [x] is it ok to have new field for `snv_pipeline_id` in the FileMetadata for the curated normals (or should the `-CURATED` suffix be added dynamically in the operator?)
- [x] clean up fixtures if needed
- [x] remove hard-coded default normal ID, use sample ID or filename
- [x] how to know which sequencer's default normal to use? -(just use NovaSeq for now)


ionox0:
I have added the Operator, as well as some fixtures for testing purposes. I'm hoping to be able to test this operator independently of the upstream pipeline by importing these fixtures, importing the necessary files into Beagle, and then triggering the operator manually.

runner/operator/access/v1_0_0/snps_and_indels/__init__.py
```diff
@@ -0,0 +1,174 @@
...
+            "location": "juno://" + tumor_bam['path']
+        }]
+
+        normal_sample_names = ['']
+        matched_normal_ids = [normal_sample_id]
+
+        # Todo: how to know which sequencer's default normal to use?
+        normal_bams = [{
+            "class": "File",
+            "location": "/ifs/work/bergerm1/ACCESS-Projects/hiseq_curated_duplex/F22_cl_aln_srt_MD_IR_FX_BR__aln_srt_IR_FX-duplex.bam"
```
ionox0, line 49: Use /work/access/production/resources or make a voyager-access-resources folder

ionox0: Also need to register in Beagle DB

ionox0: Also needs to say `juno:///`

runner/operator/access/v1_0_0/snps_and_indels/__init__.py
```diff
@@ -0,0 +1,174 @@
...
+            'igocomplete': True
+        }
+    )
+    normal_bams = [{'class': 'File', 'location': b.path} for b in normal_bams]
+    normal_ids = [n['metadata']['cmoSampleId'] for n in normal_bams]
+    return normal_bams, normal_ids
+
+
+def construct_sample_inputs(patient_id, tumor_bam, tumor_simplex_bam, tumor_sample_id, matched_normal_bam, normal_sample_id):
+    with open('runner/operator/access/v1_0_0/snps_and_indels/input_template.json.jinja2') as file:
```
nikhil, line 34: It might be better to use something like this to pull the code path: https://github.com/mskcc/beagle/blob/master/runner/operator/argos_operator/v1_1_0/construct_argos_pair.py#L15

runner/operator/access/v1_0_0/snps_and_indels/__init__.py
ionox0: Also use FileRepository.filter() instead of hard-coding

runner/operator/access/v1_0_0/snps_and_indels/__init__.py
```diff
@@ -0,0 +1,174 @@
...
+        )
+
+        sample_input = json.loads(input_file)
+        return sample_input
+
+
+class AccessLegacySNVOperator(Operator):
+    def get_jobs(self):
+
+        run_ids = self.run_ids
```
ionox0, line 89: Instead of using the run_id, filter FileRepo for all tumor bams with no associated run through variant calling

That way we can run these bams from the previous pipeline, but also ones that may have missed being run due to previously having no paired normal

runner/tests/views/test_run_api_view.py
```diff
@@ -71,13 +71,13 @@ def test_tag(self):
         url = self.api_root + '?tags=tag:value'
         response = self.client.get(url)
         self.assertEqual(response.status_code, status.HTTP_200_OK)
-        self.assertEqual(response.json()['count'], 2)
```
ionox0, line 74: @sivkovic @nikhil @stevekm 

Because I've added an extra run, pipeline, operator, and operator run to the common fixtures, these numbers in the api view test have gone up, is this ok or is there a better way to do this?

runner/operator/access/v1_0_0/snps_and_indels/test/fixtures/curated_normals_file_group.json
```diff
@@ -0,0 +1,13 @@
+[
+  {
+      "model": "file_system.filegroup",
+      "pk": "a66f03b4-ce43-40d0-9134-d5a3575cabcc",
+      "fields": {
+          "created_date": "2019-11-05T20:13:29.923Z",
+          "modified_date": "2019-11-05T20:13:29.923Z",
+          "name": "access_curated_normals",
+          "slug": "access_curated_normals",
+          "storage": "0cd14133-0f6a-413d-a498-495bf48546ee"
```
ionox0, line 10: Try linking to the common file storage fixture here, then we shouldn't need to add a new fixture for file storage

runner/operator/access/v1_0_0/snps_and_indels/test/fixtures/curated_normals_file_storage.json
```diff
@@ -0,0 +1,12 @@
+[
```
ionox0, line 1: see if we can get rid of this by using the common file storage fixture

runner/operator/access/v1_0_0/snps_and_indels/test/fixtures/files_metadata.json
```diff
@@ -0,0 +1,249 @@
...
+        "model": "file_system.filemetadata",
+        "pk": "fffff52a-0696-48d4-9ca5-6580daa44ab2",
+        "fields": {
+            "latest": true,
+            "created_date": "2020-01-06T19:59:50.583Z",
+            "modified_date": "2020-01-06T19:59:50.583Z",
+            "file": "fffff52a-0696-48d4-9ca5-6580daa44ab3",
+            "version": 0,
+            "metadata": {
+                "R": "R1",
```
ionox0, line 12: Lets see if we can get away with only having the metadata that we need for the test here, to make this pr smaller

stevekm:
all tests passed for me with `make test` on aa7957409756f47c33636323da0578c84cbc5ed9

runner/tests/views/test_run_api_view.py
stevekm: looks fine to me

stevekm:
I just saw that your test case is located at;

```
runner/operator/access/v1_0_0/snps_and_indels/test/test_snps_and_indels_operator.py
```

I think we had been putting test cases under `runner/tests` instead, not sure if it should be moved or not.

runner/operator/access/v1_0_0/snps_and_indels/test/test_snps_and_indels_operator.py
```diff
@@ -0,0 +1,99 @@
...
+    "curated_normals_file_group.json",
+    "curated_normals_file_metadata.json",
+    "files.json",
+    "files_metadata.json",
+    "operator_run.json",
+    "ports.json",
+    "runs.json",
+]
+
+# Todo: how to shorten these to remove folder names like other tests have it?
```
stevekm, line 26: I think its because you have them in the wrong place in the directory tree

Fixtures that are global for Beagle such as File Groups and File Types can be moved up to the other fixtures for those in the root of the app dir (`file_system/fixtures`) and added to the JSON's there. When the fixtures are in that location it should be able to load automatically without the full file path like the other items in the `fixtures` attribute on the test case class.

Fixtures that are sample-specific such as Runs, Ports, Files, FileMetadata, could be included in the project root fixtures dir (`beagle/fixtures`) but I dont think there was much organization or consistent labeling of those so you would be free to set up the dir for your ACCESS fixtures however you like. Note that you would still have to resolve the full path to them when you load them in the test case though, I think.

ionox0:
> I just saw that your test case is located at;
> 
> ```
> runner/operator/access/v1_0_0/snps_and_indels/test/test_snps_and_indels_operator.py
> ```
> 
> I think we had been putting test cases under `runner/tests` instead, not sure if it should be moved or not.

Yes will move them to the right place

runner/operator/access/v1_0_0/snps_and_indels/test/test_snps_and_indels_operator.py
ionox0: I've moved these fixtures to the `beagle/fixtures` dir, and removed the FileGroup fixture file in favor of adding a FileGroup to the common fixtures file. So I believe this should be resolved now!

ionox0:
@stevekm I've moved the tests and fixtures

stevekm:
all tests work for me with `make test` on 040cb5

stevekm (approved):

allanbolipata (approved):

PR #599, opened 2020-11-16
Add Operator for ACCESS SV pipeline, + accomodate changes in SNV test…

ionox0:
…s and fixtures

PR #586, opened 2020-11-06
fixes for ACCESS SNV operator

ionox0:
I had to make a few more fixes than intended, but after these changes my test run has completed successfully on staging, although still waiting on the permissions fix for final copying of outputs

PR #602, opened 2020-11-18
Access Legacy MSI Operator

aef-:


runner/operator/access/v1_0_0/msi/__init__.py
```diff
@@ -0,0 +1,145 @@
...
+from file_system.repository.file_repository import File, FileRepository
+
+
+WORKDIR = os.path.dirname(os.path.abspath(__file__))
+
+class AccessLegacyMSIOperator(Operator):
+
+    # Will be populated after get_sample_inputs()
+    sample_inputs = None
+    number_of_inputs = None
```
ionox0, line 23: I know I did it this way at first, but actually Beagle will only call `get_jobs()` so it is better to have `get_jobs()` call `self.get_sample_inputs()` and for these two variables to be returned directly. 

https://github.com/mskcc/beagle/pull/586/files#diff-0abd8d465426319d5adefec646766875f8bf4d6e526c510cb295b1233c60a333R129

Lmk it that makes sense

runner/operator/access/v1_0_0/msi/__init__.py
```diff
@@ -0,0 +1,145 @@
...
+        """
+        access_duplex_output_ports = Port.objects.filter(
+            name='duplex_bams',
+            run__app__name='access legacy',
+            run__status=RunStatus.COMPLETED
+        )
+        # Each port is a list, so need a double list comprehension here
+        all_access_output_records = [f for p in access_duplex_output_ports for f in p.value]
+        # these are port objects, they dont have metadata field
+        all_access_completed_samples = [r['sampleId'] for r in all_access_output_records]
```
ionox0, line 39: I'll try to find an example of the collapsing pipeline outputs and how the ports for that look after a completed run, so you can understand how these "record" objects work

ionox0: @aef- So, instead of this^, you can refer to how I'm doing it here:

https://github.com/mskcc/beagle/pull/599/files#diff-7f514a1f5bc70172b2c82eb7b6da8b76dc955eb056b917269abd4164bc25bb4eR28

ionox0: I'm giving up on the record outputs because I realized they are showing up as "null", so it probably makes sense to do it this way, and just use "-L0" and "_cl_aln" from the file names to get the T/N and sample ID information

ionox0:
@aef- Hey I made a few fixes here, and changed the base to the develop branch. I was able to get these test fixtures to submit successfully to staging, and then got this error from ridgeback which I'll be looking into:

```
'msisensor' not found: [Errno 2] No such file or directory: 'msisensor': 'msisensor'
ERROR:cwltool:'msisensor' not found: [Errno 2] No such file or directory: 'msisensor': 'msisensor'
[job msisensor.cwl] completed permanentFail
WARNING:cwltool:[job msisensor.cwl] completed permanentFail
```

I think we just have to make sure the PATH is also getting set properly for this pipeline, `/opt/common/CentOS_6-dev/bin/current/msisensor` is the location of the 0.2 version we use

ionox0:
Ok, just had to copy the msisensor bin, and then had to make a python 3 fix to the `admie_analyze` script. This PR LGTM now

ionox0 (approved):

PR #609, opened 2020-11-30
Ij/access cnv operator

ionox0:


ionox0:
One thing I think I may have been doing wrong is that I've been manually creating the pipeline / operator objects through the admin in staging. But maybe I'm supposed to use a migration for those things?

If anyone has a suggestion about that it could be useful for next time

fixtures/tests/access_snv/files_metadata.json
```diff
@@ -122,6 +122,7 @@
                 "labHeadName": "-",
                 "labHeadEmail": "-",
                 "piEmail": "-",
+                "sex": "M",
```
ionox0, line 125: Note that this "sex" field is required for us to be able to run the CNV pipeline, as it contains logic based on the sex of the patient.

runner/fixtures/runner.pipeline.json
```diff
@@ -132,12 +132,28 @@
             "modified_date": "2019-12-05T01:12:39.854Z",
             "name": "access legacy msi",
             "github": "git@github.com:mskcc/ACCESS-Pipeline",
-            "version": "master",
+            "version": "ij/output_bam_files_instead_of_directory",
```
ionox0, line 135: I should probably rename this branch to something like "beagle_compat", as it now has more than just the single compatibility change for getting it to work with beagle

aef- (approved):

PR #612, opened 2020-12-02
Use old CMO sample ID format for ACCESS

ionox0:
@allanbolipata I believe this will solve our issue for now, as it seems the previous format that begins with `C-` still exists in the FileMetadata

If you have an issue with us using the `sampleName` field please let us know though :)

### DIFF
--- a/runner/operator/access/v1_0_0/legacy/__init__.py
+++ b/runner/operator/access/v1_0_0/legacy/__init__.py
@@ -16,7 +16,7 @@ import json
 from jinja2 import Template
 
 REQUIRED_META_FIELDS = [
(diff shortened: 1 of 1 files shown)


aef- (approved):

allanbolipata:
@ionox0 As we discussed earlier, I have no issue with you using whichever you think is best, and it's entirely your team's decision to make. I was only explaining the reason why the IDs are formatted differently even though they represent the same sample; hopefully it helped you choose which route to take.

We will probably have issues in the future with resolving these, but I'm sure you'll be involved in crossing that bridge when we get to it.

PR #631, opened 2020-12-10
[Access] Merge FASTQ Operator

aef-:


runner/operator/access/v1_0_0/merge_fastqs/__init__.py
```diff
@@ -0,0 +1,95 @@
...
+
+def group_by_fastq(samples):
+    fastqs = defaultdict(list)
+    for sample in samples:
+        if '_R2_' in sample["path"]:
+            fastqs["R2"].append(sample)
+        else:
+            fastqs["R1"].append(sample)
+
+    return fastqs
```
aef-, line 94: What do you think of this function @ionox0 ? Should I change how I'm grouping R1s and R2s?

runner/fixtures/runner.pipeline.json
```diff
@@ -126,17 +126,17 @@
...
+            "name": "merge fastqs",
+            "github": "git@github.com:msk-access/cwl-commandlinetools",
             "version": "master",
-            "entrypoint": "workflows/subworkflows/msi.cwl",
+            "entrypoint": "merge_fastq_0.1.7/merge_fastq_0.1.7.cwl",
             "output_file_group": "a975f490-1b02-4575-abae-a4f8e3667733",
-            "output_directory": "/work/access/production/runs/voyager/microsatellite_instability",
-            "operator": 7,
+            "output_directory": "/work/access/production/runs/voyager/merged_fastqs",
+            "operator": 8,
```
aef-, line 139: I'll fix this...

runner/operator/access/v1_0_0/merge_fastqs/__init__.py
```diff
@@ -0,0 +1,95 @@
...
+        number_of_inputs = len(inputs)
+
+        return [
+            (
+                APIRunCreateSerializer(
+                    data={
+                        'name': "LEGACY FASTQ Merge: %s, %i of %i" % (self.request_id, i + 1, number_of_inputs),
+                        'app': self.get_pipeline_id(),
+                        'output_metadata': {
+
```
aef-, line 68: ITW

fixtures/tests/merge_fastqs/10151_F_13.file.json
```diff
@@ -0,0 +1,54 @@
+[
```
aef-, line 1: Currently, the test only checks for when merging is required. I made a PR to export File and File Metadata fixtures from Django Admin, when that gets merged in I'll add cases for other samples in this request.

runner/operator/access/v1_0_0/merge_fastqs/__init__.py
```diff
@@ -0,0 +1,111 @@
...
+        number_of_inputs = len(inputs)
+
+        return [
+            (
+                APIRunCreateSerializer(
+                    data={
+                        'name': "LEGACY FASTQ Merge: %s, %i of %i" % (self.request_id, i + 1, number_of_inputs),
+                        'app': self.get_pipeline_id(),
+                        'output_metadata': {key: metadata[key] for key in METADATA_OUTPUT_FIELDS if
+                                            key in metadata},
```
ionox0, line 85: I'm not sure if you were planning to remove the `igoComplete: true` filter from AccessLegacyOperator, but if not then we need to pass that field here as well

runner/operator/access/v1_0_0/merge_fastqs/__init__.py
ionox0: One thing I want to check is whether the order of these lists matters in the merge_fastq tool. I can ask Ronak

runner/operator/access/v1_0_0/merge_fastqs/__init__.py
```diff
@@ -0,0 +1,111 @@
...
+            "class": "File",
+            "location": "juno://" + fastq["path"]
+        } for fastq in fastqs["R2"]]
+
+        input_file = template.render(
+            fastq1_files=json.dumps(fastq1s),
+            fastq2_files=json.dumps(fastq2s),
+        )
+
+        inputs.append((json.loads(input_file), sample_files[0]["metadata"]))
```
ionox0, line 56: The problem with using `sample_files[0]` here is that there can be different sample ID info for the two pairs of fastqs (because they came from different sequencing runs)

For example these fields might be different between [0] and [1]:
```
    'dnaInputNg',
    'captureInputNg',
    'libraryVolume',
    'libraryConcentrationNgul'
```

In the `lims_client` I made some logic to take the average of these values:

https://github.mskcc.org/MSK-ACCESS/lims_client/blob/master/lims_client/models.py#L128-L151

But it should be ok for now unless Ronak has any concerns

fixtures/tests/merge_fastqs/10151_F_13.file.json
ionox0: 👍

runner/operator/access/v1_0_0/merge_fastqs/__init__.py
aef-: I think we can remove it. Since we're using it here, all the files are passed are `igoComplete`. It just means we have to use `self.run_ids` for the BAM generation, which I think is for the best anyways. But  let me know if you disagree.

runner/operator/access/v1_0_0/merge_fastqs/__init__.py
aef-: Ah I forgot about that. I'll average them out.

runner/operator/access/v1_0_0/merge_fastqs/__init__.py
```diff
@@ -0,0 +1,123 @@
...
+                        'tags': {'requestId': self.request_id, 'sampleId': metadata["sampleId"]}}
+                ),
+                job
+             )
+
+            for i, (job, metadata) in enumerate(inputs)
+        ]
+
+def calc_avg(sample_files, field):
+    return sum([float(s["metadata"][field] or 0) for s in sample_files if field in s["metadata"]])/float(len(sample_files))
```
ionox0, line 106: Maybe use the number of available instances in the denominator:
`field_count = len(list(filter(lambda s: s["metadata"][field], sample_files)))`
...
`... / float(len(field_count)`

runner/operator/access/v1_0_0/merge_fastqs/__init__.py
ionox0: I also think the merge_fastq required these lists to be sorted in matching order. So add this here:
```
# Sort the lists to ensure R1 and R2 fastqs are paired together
samples.sort(key = lambda s: s["path"].split("/")[-1])
```

ionox0: (on line 121)

runner/tests/operator/merge_fastqs/test_fastq_merge_operator.py
```diff
@@ -0,0 +1,66 @@
...
+
+        self.assertEqual(len(jobs) > 0, True)
+        for job in jobs:
+            self.assertEqual(job[0].is_valid(), True)
+            # Ensure at least 3 of the metadata fields are set, as not all of them are request.
+            self.assertEqual(len(job[0].initial_data["output_metadata"].keys()) > 3, True)
+            job[0].initial_data["output_metadata"]["sampleOrigin"] = "Tissue"
+            self.assertEqual(float(job[0].initial_data["output_metadata"]["captureConcentrationNm"])
+                             > 0, True)
+            self.assertEqual(job[0].initial_data["output_metadata"]["captureConcentrationNm"],
```
ionox0, line 59: 👍

runner/operator/access/v1_0_0/merge_fastqs/input_template.json.jinja2
```diff
@@ -0,0 +1,10 @@
+{
+   "fastq1": {{fastq1_files}},
```
ionox0, line 2: I'm not sure I've actually tried using `null` here, but I'm assuming that it works

ionox0 (approved):
As far as I can tell this should work 👍 

So we can merge once our test with `10151_F` has passed

PR #632, opened 2020-12-11
fix access downstream operators

ionox0:
I thought that FileMetadata objects would be available to query for files that are outputs from upstream pipeline runs, but it is just the File objects that are created.

This PR makes use of the file paths to query for matching T/N duplex/simplex/standard access bams in the ACCESS SV, SNV, and MSI operators.

Also, I added a test case for the MSI operator

### DIFF
--- a/fixtures/tests/access_snv/files_metadata.json
+++ b/fixtures/tests/access_snv/files_metadata.json
@@ -8,28 +8,7 @@
             "modified_date": "2020-01-06T19:59:50.583Z",
             "file": "fffff52a-0696-48d4-9ca5-6580daa44ab3",
             "version": 0,
-            "metadata": {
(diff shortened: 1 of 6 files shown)


fixtures/tests/access_snv/files_metadata.json
```diff
@@ -8,28 +8,7 @@
             "modified_date": "2020-01-06T19:59:50.583Z",
             "file": "fffff52a-0696-48d4-9ca5-6580daa44ab3",
             "version": 0,
-            "metadata": {
```
ionox0, line 11: I deleted this metadata to reflect the fact that it won't be present after the bam generation pipeline finishes

allanbolipata (approved):
Looks good to me.

PR #634, opened 2020-12-14
Ij/access bam search fix

ionox0:
I've encountered an error whereby QC files such as `C-DDK2LJ-L003-d_cl_aln_srt_MD_IR_FX_BR__aln_srt_IR_FX-duplex.bam.fragment-sizes` are found instead of the bam files, so I've included the `$` terminator in the file path_regex to avoid this issue.

### DIFF
--- a/runner/operator/access/v1_0_0/msi/__init__.py
+++ b/runner/operator/access/v1_0_0/msi/__init__.py
@@ -62,7 +62,7 @@ class AccessLegacyMSIOperator(Operator):
 
         for i, tumor_sample_id in enumerate(sample_ids_to_run):
             # Find the Tumor Standard bam
-            sample_regex = r'{}_cl_aln_srt_MD_IR_FX_BR.bam'.format(tumor_sample_id)
+            sample_regex = r'{}_cl_aln_srt_MD_IR_FX_BR.bam$'.format(tumor_sample_id)
             tumor_bam = FileRepository.filter(file_name_regex=sample_regex)
 
             if not len(tumor_bam) == 1:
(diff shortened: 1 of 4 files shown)


runner/operator/access/v1_0_0/structural_variants/__init__.py
```diff
@@ -15,7 +15,7 @@
 TUMOR_OR_NORMAL_SEARCH = '-L0'
 SAMPLE_ID_SEP = '_cl_aln'
 ACCESS_DEFAULT_SV_NORMAL_ID = 'DONOR22-TP'
-ACCESS_DEFAULT_SV_NORMAL_FILENAME = 'DONOR22-TP_cl_aln_srt_MD_IR_FX_BR.bam'
+ACCESS_DEFAULT_SV_NORMAL_FILENAME = 'DONOR22-TP_cl_aln_srt_MD_IR_FX_BR.bam$'
```
aef-, line 18: does this have to be regex, prepended with `r` like the others?

runner/operator/access/v1_0_0/structural_variants/__init__.py
```diff
@@ -104,7 +105,7 @@ def construct_sample_inputs(self, tumor_sample_id, tumor_bam):
             tumor_sample_names = [tumor_sample_id]
             tumor_bams = [{
                 "class": "File",
-                "location": 'juno://' + tumor_bam['path']
+                "location": tumor_bam['location']
             }]
```
ionox0, line 109: I feel like this could be a bit cleaner, for example maybe instead of re-building the file objects to put in the inputs, I could just re-use the Files directly?

Or even better, maybe there is a cleaner way to apply the port objects directly in the inputs

ionox0:
Also, since a lot of this logic is shared between the operators, it might be better to have it in a shared module

runner/operator/access/v1_0_0/structural_variants/__init__.py
ionox0: I didn't actually have the issue with this file but yes I've made it a regex as well to be sure

runner/operator/access/v1_0_0/msi/__init__.py
```diff
@@ -36,6 +36,7 @@ def get_sample_inputs(self):
         # Get the latest completed runs for the given request ID
         group_id = Run.objects.filter(
             tags__requestId=self.request_id,
+            app__name='access legacy',
```
ionox0, line 39: Had to add this line so that if the most recent run for a request ID was any pipeline besides the `access legacy` pipeline, it would find the wrong run

aef- (approved):

PR #652, opened 2020-12-28
Fix for Matched Normal bam search based on patient_id

ionox0:


runner/operator/access/v1_0_0/structural_variants/__init__.py
```diff
@@ -104,7 +104,7 @@ def construct_sample_inputs(self, tumor_sample_id, tumor_bam):
             tumor_sample_names = [tumor_sample_id]
             tumor_bams = [{
                 "class": "File",
-                "location": 'juno://' + tumor_bam['path']
+                "location": tumor_bam['location'].replace('file://', 'juno://')
```
aef-, line 107: I'm surprised this is required

aef- (approved):

runner/operator/access/v1_0_0/structural_variants/__init__.py
ionox0: Yes, I do see it being done here:

https://github.com/mskcc/beagle/blob/master/runner/operator/argos_qc_operator/v1_1_0/construct_argos_qc_outputs.py#L67

https://github.com/mskcc/beagle/blob/master/runner/operator/argos_qc_operator/v1_1_0/construct_argos_qc_outputs.py#L56

So maybe @allanbolipata or @sivkovic can comment

I'd be happy to take a look through the code to see if we can consolidate the `file` vs `juno` usage somehow

PR #655, opened 2021-01-03
Add job group to runs API endpoint

aef-:
Also adds the ability to pull pipeline by passing `/pipelines/?name=access legacy`

### DIFF
--- a/runner/views/pipeline_view.py
+++ b/runner/views/pipeline_view.py
@@ -19,18 +19,13 @@ class PipelineViewSet(mixins.ListModelMixin,
                       GenericViewSet):
     queryset = Pipeline.objects.all()
     serializer_class = PipelineSerializer
-    lookup_fields = ['pk', 'name']
 
-    def get_object(self):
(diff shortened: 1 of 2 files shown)


runner/views/pipeline_view.py
```diff
@@ -16,9 +17,16 @@ class PipelineViewSet(mixins.ListModelMixin,
                       mixins.CreateModelMixin,
                       mixins.RetrieveModelMixin,
                       GenericViewSet):
-    queryset = Pipeline.objects.order_by('id').all()
+    queryset = Pipeline.objects.all()
```
sivkovic, line 20: Should we keep some kind of ordering here?

aef-: It's your call, I didn't see a point, but without it objects are not guaranteed to come from the DB in the same order on each request.

sivkovic: You are right. Ordering for pipelines doesn't actually make a lot of sense. If we need it in the future we can easily add it.

sivkovic (approved):

PR #663, opened 2021-01-11
Feature/add commandline status

nikhil:


runner/tasks.py
```diff
@@ -488,6 +488,18 @@ def abort_job(run):
...
+    job_status_obj = {}
+    for single_commandline_job in commandline_tool_job_set:
+        status = single_commandline_job.pop('status')
+        root = single_commandline_job.pop('root')
+        if status not in job_status_obj:
+            job_status_obj[status] = []
+        job_status_obj[status].append(single_commandline_job)
+
+    run.job_statuses = job_status_obj
+    run.save()
```
sivkovic, line 501: It is always better to have one save (INSERT) statement in DB. This way you invoke two updates for each Run object for each status check.

sivkovic (approved):

PR #677, opened 2021-01-20
Updates run/restart api so as to trigger restarts by group_id OR run_id

aef-:
This PR will require one of the following parameters:

Screen Shot 2021-01-20 at 13 29 08

### DIFF
--- a/runner/views/run_api_view.py
+++ b/runner/views/run_api_view.py
@@ -201,7 +201,7 @@ class RunApiRestartViewSet(GenericAPIView):
         group_id = serializer.validated_data.get('group_id')
         if run_id:
             run = get_object_or_404(Run, pk=run_id)
-            runs = RunObject.from_db(run_id)
+            runs = [RunObject.from_db(run_id)]
         elif group_id:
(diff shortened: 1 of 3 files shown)


runner/views/run_api_view.py
```diff
@@ -193,26 +193,44 @@ class RunApiRestartViewSet(GenericAPIView):
...
+        if not serializer.is_valid():
+            return Response(serializer.errors, status=status.HTTP_400_BAD_REQUEST)
+
+        run_id = serializer.validated_data.get('run_id')
+        group_id = serializer.validated_data.get('group_id')
+        if run_id:
+            run = get_object_or_404(Run, pk=run_id)
+            runs = RunObject.from_db(run_id)
+        elif group_id:
+            runs = [RunObject.from_db(r.id) for r in Run.objects.filter(job_group_id=group_id).all()]
```
aef-, line 206: I don't fully understand `from_db` -- let me know if there is a better way to do this

sivkovic (approved):

PR #685, opened 2021-01-25
Add version/default column to pipeline admin page

aef-:
To differentiate pipelines with the same name

### DIFF
--- a/runner/views/run_api_view.py
+++ b/runner/views/run_api_view.py
@@ -281,10 +281,14 @@ class RequestOperatorViewSet(GenericAPIView):
     def post(self, request):
         request_ids = request.data.get('request_ids')
         pipeline_name = request.data.get('pipeline')
+        pipeline_version = request.data.get('pipeline_version', None)
         job_group_id = request.data.get('job_group_id', None)
         for_each = request.data.get('for_each', True)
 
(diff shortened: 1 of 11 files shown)


runner/views/run_api_view.py
```diff
@@ -281,10 +281,14 @@ class RequestOperatorViewSet(GenericAPIView):
...
         pipeline_name = request.data.get('pipeline')
+        pipeline_version = request.data.get('pipeline_version', None)
         job_group_id = request.data.get('job_group_id', None)
         for_each = request.data.get('for_each', True)
 
-        pipeline = get_object_or_404(Pipeline, name=pipeline_name)
+        if pipeline_version:
+            pipeline = get_object_or_404(Pipeline, name=pipeline_name, version=pipeline_version)
+        else:
+            pipeline = get_object_or_404(Pipeline, name=pipeline_name)
```
aef-, line 291: @sivkovic may have some input but this will error out if multiple objects are returned, we may want to add `default=True` here

ionox0: @aef- yes made that change

I noticed that it's possible for two similar pipelines to both be set as "default", but maybe that is ok

ionox0 (approved):

PR #691, opened 2021-01-27
Batch Job status calls

aef-:
Dependent on: https://github.com/mskcc/ridgeback/pull/137

We should deploy Ridgeback with that PR, then this one.


nikhil (approved):

aef-:
I'd like to chunk query results so that we're sending batches of 900 or so to Ridgeback

PR #696, opened 2021-02-02
Release/1.36.0

nikhil:
- Bump version to 1.36 (#695)
- Remove pipeline name requirement (#694)
- Added pipeline version to swagger (#690)
- Remove DB_CONN_MAX_AGE (#689)
- Use file path for sample id retrieval for ACCESS Operator (#692)
- Sort keys when querying for multiple value_lists (#669)
- Argos pairing refactor (#697)

PR #698, opened 2021-02-02
copy files

nikhil:


allanbolipata (approved):

allanbolipata:
Missed this commit into #696 , will just add release along with it

PR #725, opened 2021-02-17
Feature/fixed psycopg2 build error

nikhil:


allanbolipata (approved):

aef-:
I'm not sure why we need both? `binary` is typically used for dev environments in case the user has issues building it.

nikhil:
Looks like the binary should not be used for production according to the developers: https://www.psycopg.org/docs/install.html

PR #731, opened 2021-02-23
Add Ridgeback DB connection pooling

aef-:
I was experiencing DB connection timeouts last week within Ridgeback -- I figured instead of creating another instance of PGBouncer we can use the same one. PGBouncer creates a separate pool for each database. I haven't tried this, and I'm not entirely sure how auth works with multiple databases. It's currently running on staging.


sivkovic (approved):

allanbolipata:
Waiting to merge this until @sivkovic and @aef- test

PR #732, opened 2021-02-24
ACCESS SNV pipeline - genotype tumors and normals from same Pool

ionox0:
ACCESS SNV pipeline should genotype a selection of tumors and normals from the same capture as is being variant called


aef- (approved):

ionox0:
@aef- Just wondering if you had any thoughts on how I can optimize the queries here:

https://github.com/mskcc/beagle/pull/732/files#diff-0abd8d465426319d5adefec646766875f8bf4d6e526c510cb295b1233c60a333R217 

I think the reason it is slow is because I am going back to the fastqs, and then generating a query by OR'ing together a list of sample IDs. 

This would be  faster if the capture ID was stored on the Bam Files directly, so maybe I can include that field when I refactor these operators to use `output_metadata`, but lmk if you happen to see another way I'm missing

ionox0:
I guess it might also be slow because of one of these
```
                .distinct('file_name')\
                .order_by('file_name', '-created_date')\
```

runner/operator/access/__init__.py
```diff
@@ -35,21 +35,50 @@ def get_request_id_runs(request_id):
...
+        unfiltered_matched_normal_bam = File.objects.filter(
+            file_name__startswith=patient_normals_search,
+            file_name__endswith=IGO_UNFILTERED_REGEX
+        )
+
+    if len(unfiltered_matched_normal_bam) > 1:
+        msg = 'WARNING: Found more than one matching unfiltered normal bam file for patient {}. ' \
+              'We will choose the most recently-created one for this run.'
+        msg = msg.format(patient_id)
+        logger.warning(msg)
```
aef-, line 74: I think for debugging it'd help to have the group_id in addition to the patient id, you can access it with `self.job_group_id`

ionox0: because this is an external function I don't have access to that field of the operator, it seems a bit messy to pass that in to this function but is there another way to get it?

PR #737, opened 2021-03-02
Send FAILED notification when creating Run fails

sivkovic:
## Changelog
- Send notification on FAIL to create Run
- Pep8 fixes


runner/tasks.py
```diff
@@ -288,10 +289,7 @@ def create_run_task(run_id, inputs, output_directory=None):
         run = RunObject.from_cwl_definition(run_id, inputs)
         run.ready()
     except RunCreateException as e:
-        run = RunObject.from_db(run_id)
-        run.fail({'details': str(e)})
-        RunObject.to_db(run)
-        logger.info("Run %s Failed" % run_id)
+        fail_job(run_id, str(e))
```
aef-, line 292: I think we should remove the try catch and let Celery take care of it here:
https://github.com/mskcc/beagle/blob/6777c856ea8396f56c50b3ef3e20393ca8770c1e/runner/tasks.py#L278

aef- (changes requested):

aef- (approved):

PR #751, opened 2021-03-12
Pull request id from previous runs when available

aef-:
Instead of having to go fishing for the `request_id` in the operator, if the first run in a chain has a `requestId` tag set, we should just pass it along to downstream operators that were triggered

### DIFF
--- a/runner/tests/test_trigger.py
+++ b/runner/tests/test_trigger.py
@@ -6,7 +6,7 @@ import uuid
 from mock import patch, call
 from django.test import TestCase
 from runner.models import OperatorRun, RunStatus, TriggerRunType, OperatorTrigger, Run
-from runner.tasks import process_triggers, complete_job, fail_job, create_jobs_from_operator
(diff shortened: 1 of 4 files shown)


runner/tests/test_trigger.py
```diff
@@ -115,9 +115,29 @@ def test_operator_trigger_executes_runs_individually(self, create_jobs_from_chai
...
                  )
         ]
-
         create_jobs_from_chaining.delay.assert_has_calls(calls, any_order=True)
         process_triggers()
         operator_run.refresh_from_db()
         self.assertEqual(operator_run.status, RunStatus.COMPLETED)
 
+    @patch('runner.tasks.create_jobs_from_operator')
+    def test_operator_trigger_passes_request_id_tag_if_one_is_not_provided(self, create_jobs_from_operator):
```
ionox0, line 124: should this be called "_if_one_is_provided" ?

aef-: i meant as in not provided to OperatorFactory, i can clarify

aef-:
Closing for https://github.com/mskcc/beagle/pull/769

PR #759, opened 2021-03-18
Add command to sync with DB on prod

aef-:
This takes the latest DB dump and loads it into your local DB. I also added a flag to the `pg_dump` command in our back-up cronjob that reduces the total size of the dump. Information on that can be viewed here: https://dan.langille.org/2013/06/10/using-compression-with-postgresqls-pg_dump/

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -174,7 +174,7 @@ db-backup: $(DB_BACKUP_DIR)
 db-restore:
 	@echo ">>> restoring db from DBFILE: $(DBFILE)"
 	if [ -n "$(DBFILE)" ]; then \
-	pg_restore -d "$(PGDATABASE)" < "$(DBFILE)" ; fi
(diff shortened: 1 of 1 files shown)


aef- (commented):
Have to add obfuscation to DMP BAM filename and path.

Makefile
```diff
@@ -114,6 +114,8 @@ install: conda
 # pip install git+https://github.com/mskcc/beagle_cli.git@develop
 # pip install -r requirements-cli.txt # <- what happened to this file?
 
+BEAGLE_BACKUP_DB_FOLDER=/juno/work/ci/beagle_backups/
```
ionox0, line 117: not sure if this just makes it more complicated than necessary but this could maybe be an env var

ionox0:
@aef- just FYI I used this locally and it worked for me!

ionox0 (approved):

Issue #762, opened 2021-03-19
Restart does not resume run


nikhil:
I tried to restart operator run: b503e4a2-4137-4f8f-8484-7f3afcfccfa0 and its run (0343affe-8c09-4005-9b3e-ed3a7c3df49a) has a valid jobstore.

```{
  "runs_restarted": [
    "b7455885-bd16-4a81-9b1a-5cb36d5bdd54"
  ],
  "runs_copied": []
}
```

but ridgeback shows the new run: (http://voyager:5003/v0/jobs/987c8df9-57f5-4570-b4b1-8de44bc2c160/)
has 
```
"resume_job_store_location": null,
```

nikhil:
The run also gives a strange error 
```
cwltool.errors.WorkflowException: ../../work/ci/beagle/work/987c8df9-57f5-4570-b4b1-8de44bc2c160/pluto-cwl/cwl/workflow_with_facets.cwl:259:3: Missing req    uired input parameter 'IMPACT_gene_list'
```

and the command does not have --restart or the original jobstore

```
cwltoil --singularity --coalesceStatusCalls --logFile toil_log.log --batchSystem lsf --disable-user-provenance --disable-host-prov    enance --stats --debug --disableCaching --preserve-environment PATH TMPDIR TOIL_LSF_ARGS SINGULARITY_PULLDIR SINGULARITY_CACHEDIR PWD SINGULARITY_DOCKER_U    SERNAME SINGULARITY_DOCKER_PASSWORD --defaultMemory 8G --maxCores 16 --maxDisk 128G --maxMemory 256G --not-strict --realTimeLogging --jobStore /work/ci/be    agle/job-store/987c8df9-57f5-4570-b4b1-8de44bc2c160 --tmpdir-prefix /work/ci/beagle/tmp/987c8df9-57f5-4570-b4b1-8de44bc2c160 --workDir /work/ci/beagle/wor    k/987c8df9-57f5-4570-b4b1-8de44bc2c160 --outdir /juno/res/ci/voyager/b7455885-bd16-4a81-9b1a-5cb36d5bdd54 --maxLocalJobs 500 /work/ci/beagle/work/987c8df9    -57f5-4570-b4b1-8de44bc2c160/pluto-cwl/cwl/workflow_with_facets.cwl /work/ci/beagle/work/987c8df9-57f5-4570-b4b1-8de44bc2c160/input.json
```

ionox0:
[...]

PR #767, opened 2021-03-22
fix restart - new run should copy ports from old run (not from itself)

ionox0:
https://github.com/mskcc/beagle/issues/762

### DIFF
--- a/runner/views/run_api_view.py
+++ b/runner/views/run_api_view.py
@@ -252,7 +252,8 @@ class RunApiRestartViewSet(GenericAPIView):
             r.save()
 
             # Copy over the input / output ports from the original Run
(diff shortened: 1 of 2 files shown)


sivkovic (approved):

runner/views/run_api_view.py
```diff
@@ -252,7 +252,8 @@ def post(self, request):
             r.save()
 
             # Copy over the input / output ports from the original Run
-            ports = r.port_set.all()
+            original_run = Run.objects.get(id=original_run_id)
+            ports = original_run.port_set.all()
```
aef-, line 256: good catch -- you could just put `ports = r.port_set.all()` below the loop instead

PR #769, opened 2021-03-23
Ensure request_id is always available within the ACCESS pipeline

aef-:


ionox0:
What was the run / request for which this failed?

aef-:
This breaks for all runs triggered with an OperatorTrigger

runner/operator/access/v1_0_0/cnv/__init__.py
```diff
@@ -36,11 +36,7 @@ def get_sample_inputs(self):
 
         :return: list of json_objects
         """
-        if self.request_id:
-            run_ids = get_request_id_runs(self.request_id)
-            run_ids = [r.id for r in run_ids]
-        else:
-            run_ids = self.run_ids
+        run_ids = self.run_ids if self.run_ids else [r.id for r in get_request_id_runs(self.request_id)]
```
ionox0, line 39: I guess I'd want to make sure that our normal triggers (manually through the /operator/runs endpoint) don't populate `self.run_ids`, since that field will take precedence after this change

ionox0: (but I'm assuming that it's undefined when triggering it manually, since there would be no upstream runs to pull from)

aef-: yeah exactly, I'll be sure though to test out this scenario on staging

ionox0:
so is the bug that there is no tag for the request ID in the downstream runs? or is there an error

aef-:
Yup -- it's the former

ionox0 (approved):

PR #772, opened 2021-03-24
Ensure that running status in Beagle means it's running in Ridgeback

aef-:
Let's leave Runs as CREATED/READY until they're actually RUNNING in Ridgeback. I'm not sure if this will have any effect on Seqosystem -- @nikhil what do you think?

Essentially I want to be able to see what Runs will finish first in Beagle, which atm is a little difficult to do.

### DIFF
--- a/runner/tasks.py
+++ b/runner/tasks.py
@@ -350,7 +350,6 @@ def submit_job(run_id, output_directory=None):
         response = requests.post(url, json=job)
     if response.status_code == 201:
         run.execution_id = response.json()['id']
-        run.status = RunStatus.RUNNING
         logger.info("Job %s successfully submitted with id:%s" % (run_id, run.execution_id))
         run.save()
     else:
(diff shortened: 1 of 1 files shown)


runner/tasks.py
```diff
@@ -505,7 +506,8 @@ def check_job_timeouts():
     TIMEOUT_BY_DAYS = 3
     diff = datetime.now()-timedelta(days=TIMEOUT_BY_DAYS)
     runs = Run.objects.filter(status__in=(RunStatus.CREATING, RunStatus.READY),
-                              created_date__lte=diff).all()
+                              created_date__lte=diff,
+                              execution_id__isnull=True).all()
```
aef-, line 510: If we have 10000 runs queued up, with 400 running in Ridgeback at a time, we want to ensure that they aren't timed out if they have an execution_id. We may be able to remove this beat and instead rely entirely on timeouts in Ridgeback and `on_failure` in `create_run_task`.

runner/tasks.py
```diff
@@ -350,7 +350,6 @@ def submit_job(run_id, output_directory=None):
         response = requests.post(url, json=job)
     if response.status_code == 201:
         run.execution_id = response.json()['id']
-        run.status = RunStatus.RUNNING
```
aef-, line 353: a run isn't actually running here, it just received an execution_id

nikhil:
I like this idea. Should we have a new status for runs submitted but not running on ridgeback like Queued? I know sometimes runs don't get submitted due to a bug and we figure it out by seeing that it is stuck in the created state.

aef-:
I figured adding another status may cause some complications. We get the QUEUE status with the following:
```
CREATED -- run was created in Beagle
READY   -- CWL/inputs verified
            -- without execution_id -- Attempting to submit to ridgeback (this status doesn't last long, either it gets an execution id or fails within a few seconds)
            -- with execution_id -- Queued for run in Ridgeback
RUNNING -- it's running on Ridgeback
```

sivkovic (approved):

PR #773, opened 2021-03-24
additional validation for access legacy operator

ionox0:
I also moved the access tests to their own subfolder to keep it a bit more organized


ionox0:
WIP I probably want to add some tests

runner/operator/access/v1_0_0/legacy/__init__.py
```diff
@@ -38,7 +55,18 @@
...
 """
 def get_missing_fields(source, fields):
-    return list(filter(lambda field: field not in source or not source[field], fields))
+
+    def validate(field):
+        if field not in source:
+            return True
+        elif not source[field]:
+            return True
+        elif type(source[field]) == list:
```
ionox0, line 64: @aef- this is to handle the failure we had where just one of the fields in the `add_rg_PU: ['IDT234', None, ...]` field was null

ionox0:
@aef- @sivkovic What does the `InputCreationFailedEvent` do? I was hoping that this would handle the notification for letting us know the inputs are invalid

aef-:
I created this error message specifically for input validation
https://github.com/mskcc/beagle/blob/master/runner/operator/access/v1_0_0/legacy/__init__.py#L132

ionox0:
Ok great, and that notifier is set up to put a comment in the Jira ticket right?

aef- (approved):

PR #782, opened 2021-03-29
Argos operator/fcid correction

allanbolipata:
Covers https://app.clickup.com/t/pk8af8 - prevents RG ID ending up empty since it requires platform unit/flowcell Id to generate one.

### DIFF
--- a/runner/operator/argos_operator/v1_1_2/bin/make_sample.py
+++ b/runner/operator/argos_operator/v1_1_2/bin/make_sample.py
@@ -265,7 +265,7 @@ class Fastqs:
         pu = list()
         for f in self.r1:
             metadata = get_file(f).metadata
-            flowcell_id = ""
+            flowcell_id = "MT_FCID"
             if 'poolednormal' in self.sample_name.lower():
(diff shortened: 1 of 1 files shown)


runner/operator/argos_operator/v1_1_2/bin/make_sample.py
```diff
@@ -265,12 +265,15 @@ def _set_pu(self):
         pu = list()
         for f in self.r1:
             metadata = get_file(f).metadata
+            flowcell_id = "MT_FCID"
```
sivkovic, line 268: I think in these cases you can always do `metadata('flowCellId', "MT_FCID")` if metadata['flowCellId'] = None. But this is good also.

sivkovic (approved):

runner/operator/argos_operator/v1_1_2/bin/make_sample.py
allanbolipata: Good point! I'll make a note of it for future releases.

PR #788, opened 2021-03-30
Update construct_argos_pair.py

allanbolipata:
Adds IMPACT410 bait set


timosong:
looks great

allanbolipata:
@timosong looks great, yeah, but i think this wouldn't work ultimately because the pairing logic for dmp bams and pooled normals actually go by the name of the bait set.

for a complete solution, we'd need to revisit that. Or the data tables should be updated to reflect the right bait set value.

closing PR for now, we will do it through the data until we have a more concrete plan

PR #790, opened 2021-04-01
Add Operator Run endpoint with better filtering for beagle_cli

aef-:


runner/views/run_api_view.py
```diff
@@ -18,19 +18,26 @@
...
 from drf_yasg.utils import swagger_auto_schema
 from beagle.common import fix_query_list
 from notifier.events import RunStartedEvent, AddPipelineToDescriptionEvent
 
+def query_from_dict(query_filter,queryset,input_list):
+    for single_input in input_list:
+        key, val = single_input.split(':')
+        query = {query_filter % key: val}
+        queryset = queryset.filter(**query).all()
+    return queryset
```
ionox0, line 40: This appears similar to `query_from_dict` in the other file but slightly different, I'd just make sure your new version works as expected

ionox0 (approved):
I'm assuming the purpose of this is so that you can query by app / app_version / tags / status from beagle_cli

If so it lgtm

PR #792, opened 2021-04-01
Add operator run param to run api

aef-:


aef-:
I'll change the base when dev and master are lined up

ionox0 (approved):

PR #813, opened 2021-04-08
update path to new gbcms 1.2.4

ionox0:


ionox0:
@aef- this is finished, I've ran the pipeline several times with these new tools installed

aef- (approved):

PR #841, opened 2021-05-13
Changed 640 -> 644 and 750 -> 751 so tempobot can access files

allanbolipata:
Tempo users on JUNO wasn't getting access to the files; had to update file permissions so that users can at least traverse through the dirs (644) and users can at least read the files


sivkovic (approved):

allanbolipata:
We don't need to merge this - the relevant parties have been added to a different group that does have permissions; I will make a new ticket for that

allanbolipata:
This needs to be reopened so that we can set the proper permissions for all the relevant groups and pipeline users while certain issues linger.

Although a new group was created so that the previous permissions would be viable for all users in that group (`cmoigo`), the primary group of `voyager` in our main production machine is not set to that. Changing group ownership of files/dirs, however, require `root` or elevated privileges, something that is not allowed on our current systems.

For now, we still need to make these fastq files available for other users, so we have to open permissions a little bit so that their operations are not affected.

A decision will have to be made in the future:

1. Do we set the primary group of `voyager` user to `cmoigo`? How would that affect our process?
2. Do we wait until an alternative `beagle` service is spun up on our recently-provisioned Docker VM, which does grant us elevated privileges (though there may still be other things to consider, since communication with LSF and the VM is still being evaluated)?
3. Do we just leave permissions as-is? 644 for dirs, 751 for files.

allanbolipata:
Nevermind, @stevekm figured it out. Closing this out in favor of better group ownership logic

PR #848, opened 2021-05-19
Operator to generate inputs for sample sheets

aef-:


ionox0:
Where does this sample sheet get used? I think IGO also makes these so it could be simpler to just import the existing ones into beagle?

runner/operator/access/v1_0_0/sample_sheet/__init__.py
```diff
@@ -0,0 +1,64 @@
...
+        samples = []
+        for f in files:
+            if f.metadata["barcodeIndex"]:
+                barcodeIndex = f.metadata["barcodeIndex"].split("-")
+                index1 = barcodeIndex[0]
+
+            for lane in f.metadata["flowCellLanes"]:
+                samples.append({
+                    "Lane": lane,
+                    "SampleID": f.metadata["cmoSampleName"][2:].replace("_", "-"),
```
ionox0, line 30: You can just use the `sampleName` field here instead of the split / replace

ionox0 (approved):
I can talk to David about the possibility of importing these sample sheet files from their end since it would probably make this simpler, but if that is not an option this might be the only way

PR #849, opened 2021-05-21
Add group owner settings, recursive group owner change in copy service

allanbolipata:
Added changing group owner to `cmoigo` after copy.

I used `black` to reformat `copy_service.py`.


beagle_etl/copy_service/copy_service.py
```diff
@@ -8,23 +8,33 @@
...
         dirname = os.path.dirname(path_to)
         if not os.path.exists(dirname):
             os.makedirs(dirname, mode=settings.COPY_DIR_PERMISSION)
         copyfile(path_from, path_to)
         os.chmod(path_to, settings.COPY_FILE_PERMISSION)
 
+        for dirpath, dirnames, filenames in os.walk(path_to):
+            shutil.chown(dirpath, group=settings.COPY_GROUP_OWNERSHIP)
+            for filename in filenames:
+                shutil.chown(
```
sivkovic, line 27: This will chown even `/juno/work/ci/` I think. Is that expected?

allanbolipata: If that's what it does, I will have to change it - it should only set permissions for files/dirs below that root dir

allanbolipata:
I was working on this.

It looks like there's no easy way to change group ownership unless it's actually done by `root`. I need to look into it some more, so I will close this for now.

Issue #917, opened 2021-08-18
Some pooled normal fastq files show up twice in the mapping file


allanbolipata:
Duplicates seem to show up for some fastq files when the mapping text file is generated.

See for example: http://jira.mskcc.org:8090/browse/VA-1209

PR #918, opened 2021-08-18
Fixed sample mapping output generator for argos operator 1.1.3

allanbolipata:
R2 pooled normals were being duplicated in mapping files; removed conflicting code.

Also adjusted tumor sample bam mapping string generator - it had `normal_sample_name` in there.


allanbolipata:
Solves #917

sivkovic (approved):

PR #981, opened 2021-12-15
Feature/ultron fr

allanbolipata:
Still working on updating tests for this, but this operator should work.

Making PR now for visibility and for working on merge conflicts.


allanbolipata:
I don't think this test is reliable - the JSON generated sometimes has a different order and thus the test will fail.

Need to find a better way to do this test.

allanbolipata:
I checked per key instead of the json so that the values can be sorted.

There is a concern that it's not checking the order in which the values in each array gets added - because of cardinality and how indices of one array is relative to another - but it should be fine for now.

@nikhil this is ready to be reviewed.

nikhil (approved):

PR #1003, opened 2022-03-04
Request, Sample and Patient get_or_create in import jobs

sivkovic:


allanbolipata (approved):

allanbolipata:
@sivkovic I think I missed this before but it looks like this is no longer needed

we can probably delete this, but can you confirm

Issue #1028, opened 2022-04-28
Need to updated paths


rhshah:
@nikhil @buehlere 
We recently closed Ian's account and did not realize he had paths from his home directory. My apologies for the trouble:

https://github.com/mskcc/beagle/blob/6b8e22c3176c960a4078eec30080e452a564bcbf/runner/operator/access/v1_0_0/legacy/input_template.json.jinja2#L37

Should be:
```bash
    "picard_path": "juno:///juno/work/access/production/resources/tools/picard/versions/v2.8.1/picard-2.8.1.jar",
```

https://github.com/mskcc/beagle/blob/6b8e22c3176c960a4078eec30080e452a564bcbf/runner/operator/access/v1_0_0/legacy/input_template.json.jinja2#L34

Should be:
```bash 
    "arrg_path": "juno:///juno/work/access/production/resources/tools/picard/versions/v1.96/picard-tools-1.96/AddOrReplaceReadGroups.jar",
```

https://github.com/mskcc/beagle/blob/6b8e22c3176c960a4078eec30080e452a564bcbf/runner/operator/access/v1_0_0/legacy/input_template.json.jinja2#L42

Should be
```bash
    "marianas_path": "juno:///juno/work/access/production/resources/tools/marianas/current/Marianas-1.8.1.jar",
```

https://github.com/mskcc/beagle/blob/6b8e22c3176c960a4078eec30080e452a564bcbf/runner/operator/access/v1_0_0/legacy/input_template.json.jinja2#L36

Should be
```bash
    "waltz_path": "juno:///juno/work/access/production/resources/tools/waltz/current/Waltz-3.2.0.jar",
```

buehlere:
Opened a [pull request ](https://github.com/mskcc/beagle/pull/1031)with the fixes!

PR #1037, opened 2022-05-02
Hotfix/access path updates

buehlere:
✔️  updating path formatting. Some path needs different formatting since it is defined as a `string input` in the cwl and not a `File input`. The ` juno://` suffix should only be used for `File input`. 

PR relates to this original issue: https://github.com/mskcc/beagle/issues/1028


runner/operator/access/v1_0_0/legacy/input_template.json.jinja2
```diff
@@ -34,7 +34,7 @@
     "arrg_path": "juno:///juno/work/access/production/resources/tools/picard/versions/v1.96/picard-tools-1.96/AddOrReplaceReadGroups.jar",
     "gatk_path": "/opt/common/CentOS_6/gatk/GenomeAnalysisTK-3.3-0/GenomeAnalysisTK.jar",
     "waltz_path": "juno:///juno/work/access/production/resources/tools/waltz/versions/v2.0.0/Waltz-2.0.jar",
```
nikhil: Looks like this PR is missing some fixes, like here

nikhil (approved):

PR #1085, opened 2022-06-27
Bugfix/get project prefix

allanbolipata:


runner/operator/argos_operator/v1_2_0/argos_operator.py
```diff
@@ -267,7 +267,7 @@ def get_files_for_pairs(self, pairing):
     def get_files(self, request_id):
         files = FileRepository.filter(
             queryset=self.files,
-            metadata={settings.REQUEST_ID_METADATA_KEY: request_id, "igocomplete": True},
+            metadata={settings.REQUEST_ID_METADATA_KEY: request_id, "igoComplete": True},
```
sivkovic, line 270: It is probably better to use igoComplete variable from settings.py https://github.com/mskcc/beagle/blob/master/beagle/settings.py#L430

runner/operator/argos_qc_operator/v1_1_0/construct_argos_qc_outputs.py
```diff
@@ -212,7 +212,7 @@ def get_output_directory_prefix(run_id_list):
 def get_project_prefix(run_id_list):
     project_prefix = set()
     for single_run_id in run_id_list:
-        project_prefix_port = Port.objects.filter(run=single_run_id, name="project_prefix")
+        project_prefix_port = Port.objects.filter(run=single_run_id, name="project_prefix")[0]
```
sivkovic, line 215: Instead of `Port.objects.filter(run=single_run_id, name="project_prefix")[0]`, using `first()` is more optimal
```
Port.objects.filter(run=single_run_id, name="project_prefix").first()
```

sivkovic (approved):

sivkovic (approved):

PR #1101, opened 2022-08-01
Tempo Nextflow Operator

sivkovic:


runner/operator/tempo_nextflow_operator/tempo_nextflow_operator.py
```diff
@@ -0,0 +1,410 @@
...
+        #     Writing files to {file_path}.
+        #
+        #     Run Date: {run_date}
+        #     Beagle Version: {beagle_version}
+        #     """.format(
+        #         file_path=self.OUTPUT_DIR, run_date=run_date, beagle_version=beagle_version
+        #     )
+        # )
+
+        return [RunCreator(**job) for job in jobs]
```
allanbolipata, line 174: I think since this is mostly still based on the old TempoMPGen operator, we're going to be submitting every WES sample in the beagle database, which will amount to like 7000+ jobs each time we execute it.

I think the framework is here, but we'll need to update this so that we're running just new WES samples.

We may need to talk about adding some of the other text files we output here - mapping, pairing, unpaired, etc. files -  into their own tables in beagle somewhere.

Let's discuss this soon.

runner/operator/tempo_nextflow_operator/tempo_nextflow_operator.py
```diff
@@ -0,0 +1,410 @@
...
+
+    def write_historical_pairing_file(self, pairing_file_str):
+        """
+        Writes file to temporary location, then registers it to the temp file group
+        Also uploads it to notifier if there is a job group id
+        """
+        output = PAIRING_FILE_LOCATION
+        with open(output, "w+") as fh:
+            fh.write(pairing_file_str)
+        os.chmod(output, 0o777)
```
allanbolipata, line 230: The output of this function may need to be separated out into its own object/database entries because Tempo team may rely on it.

runner/operator/tempo_nextflow_operator/tempo_nextflow_operator.py
```diff
@@ -0,0 +1,410 @@
...
+                        running.append(sample.cmo_sample_name)
+                        for key in key_order:
+                            v = meta[key]
+                            s = self._validate_to_str(v)
+                            running.append(s)
+                        for key in extra_keys:
+                            s = self._validate_to_str(meta[key])
+                            running.append(s)
+                        tracker += "\t".join(running) + "\n"
+        return self.write_to_file("sample_tracker.txt", tracker)
```
allanbolipata, line 410: This tracker file was being used by Tempo team to populate the CCS' WES tracker HERA, used by the PMs.

We need to check with CCS to see if this is going to need continued support in the future as I believe HERA is at or approaching end-of-life.

allanbolipata (changes requested):
I think this is going to overload our system, unless I missed something.

I think this operator is mostly there - we would just change the inputs from "all WES samples in beagle DB" to "new samples recently received from SMILE" - but we will need to discuss other components that's going to have to be built.

allanbolipata (approved):

PR #1106, opened 2022-08-23
Feature/make m1

buehlere:
@alyvann and I were having some trouble setting up dev instances on our m1 macs. We realized this is because some packages used by `beagle` are only available on x86 macs such as ncurses. This was causing package conflicts since miniconda was mixing m1 and x86 packages together. See this post here: https://stackoverflow.com/questions/65415996/how-to-specify-the-architecture-or-platform-for-a-new-conda-environment-apple. 

This pull request incorporates the solution from the link above into the makefile along with some edits to the documentation. 

@nikhil let me know how things look / what needs to be changed. Also, @alyvann would you mind trying the solution on your machine? Thanks for the help!

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,5 +1,6 @@
 SHELL:=/bin/bash
 UNAME:=$(shell uname)
+UNAME_P:=$(shell UNAME)_$(shell uname -p)
 CURDIR_BASE:=$(shell basename "$(CURDIR)")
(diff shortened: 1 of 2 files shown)


alyvann:
The new `make install-m1` works well on my machine. Thank you, @buehlere

nikhil (approved):

nikhil:
This is great! Thanks for this!

PR #1117, opened 2022-09-21
Feature/nucleo chaining

buehlere:
✔️ handling `self.request_id` null case in nucleo_qc operator. This is causes an issue where the operator won't chain correctly.


nikhil (approved):
This should fix it!

PR #1118, opened 2022-09-21
Feature/nucleo qc agg

buehlere:
✔️ adding nucleo_qc_aggregate operator. This operator is meant to run a workflow that aggreges runs from the nucleo_qc operator. 
✔️ adding test for nucleo_qc_aggreagate operator.


runner/operator/cmo_ch/v2_1_0/qc_agg/nucleo_qc_operator_agg.py
```diff
@@ -0,0 +1,189 @@
...
+                    "name": "CMO-CH Nucleo QC: %s, %i of %i" % (self.request_id, i + 1, len(sample_inputs)),
+                    "app": self.get_pipeline_id(),
+                    "inputs": job,
+                    "tags": {settings.REQUEST_ID_METADATA_KEY: self.request_id, "cmoSampleId": job["sample_name"]},
+                }
+            )
+            for i, job in enumerate(sample_inputs)
+        ]
+
+    def get_nucleo_outputs(self):
```
nikhil, line 76: Could we modify this method in the similar way we did here: https://github.com/mskcc/beagle/pull/1116 (ie adding the case when runs are passed in the chaining process)

nikhil (changes requested):

runner/operator/cmo_ch/v2_1_0/qc_agg/nucleo_qc_operator_agg.py
buehlere: Just added the fix!

nikhil (approved):

PR #1128, opened 2022-10-17
Get run objects from run id list

nikhil:


buehlere (approved):
Looks good, successfully transforms the list of ids to query objects!

PR #1150, opened 2022-12-09
Helix Filters operator v22 update

allanbolipata:
This is an update to the helix filters operator

Main thing it adds is support for creating data clinical files for DMP tumor BAM files. These files allow helix filters to run properly.

It's only been tested on delphi; this will require a updated test datasets - one with DMP Tumor BAMs only and another with a mix of DMP Tumor BAMs and regular SMILE sample metadata.

We won't deploy these into production until tests complete, of course; I'll put the above TODO tasks on clickup for an upcoming sprint.


beagle/__init__.py
```diff
@@ -1 +1 @@
-__version__ = "1.69.1"
+__version__ = "1.70.0"
```
nikhil, line 1: You will probably need to update the version so its at or above this release: https://github.com/mskcc/beagle/pull/1172

nikhil (changes requested):

nikhil (approved):

PR #1152, opened 2022-12-20
Feature/handle image cache

nikhil:
Still needs:
- [ ] Unit tests
- [ ] Testing on staging


nikhil:
Closing since image cache was added in: https://github.com/mskcc/ridgeback/pull/292

PR #1158, opened 2023-01-05
Added resume and restart logic

nikhil:


allanbolipata (commented):
What happens if this function returns None?

runner/models.py
```diff
@@ -214,29 +215,40 @@ def clear(self):
...
+            if "restart" not in message:
+                message["restart"] = []
+            message["restart"].append((started_strftime, exited_strftime))
+            self.clear().save()
+            self.restart_attempts = restart_attempts
+            self.message = message
+            self.output_directory = output_directory
+            self.save()
+        else:
+            return None
```
allanbolipata, line 251: This function, I mean. What happens if it returns None?

nikhil: Returning None signifies that the run is out of retries and resumes and should fail. The fail task has the logic here to read None and proceed with failing the run.

https://github.com/mskcc/beagle/blob/master/runner/tasks.py#L515

allanbolipata (approved):

PR #1190, opened 2023-04-10
Improve email notifications

sivkovic:


nikhil (approved):
Looks good, I think the only thing is fixing the line break like we discussed this morning.

PR #1214, opened 2023-05-22
Fix/invalid pipeline operator factory

allanbolipata:
I think there's a disconnect between the version of the operator and the version of the pipeline.

Based on the test comment here, I'm implementing the fix for that.

https://github.com/mskcc/beagle/blob/5452f9ebd36b3d316c40707f1021d34e497c77a7/runner/tests/operator/test_operator_factory.py#L26-L30


allanbolipata:
Yeah let's not do this. 

Although the commits here fixes the test, I don't think it's a necessary test. @nikhil you can reopen if you think it's necessary - since you added this test a long time ago - but the OperatorFactory has changed so much since then that this seems outdated.

Closing.

PR #1227, opened 2023-05-26
release/1.77.2

allanbolipata:
changelog
---
Adds a hotfix for balto operator:
- We will now provide the full oncokb rds file path to the CWL as input
- Operator registers this file if it isn't in the db
- `sample_ids` input now only contains non-normal samples, so long as  `Sample.sample_type` does not contain the value `Normal`
- Tests included


allanbolipata:
Inputs aren't being created for some reason in the Run object, so these are stuck at CREATING step.

As far as I can see, there seems to be a silent error in the RunCreator that I need to debug. Tests create properly formatted JSON, though.

allanbolipata:
This was a URI assignment issue. Seems like it's running now; will merge after review and confirmation of multiple completed runs.

nikhil (approved):

PR #1233, opened 2023-06-07
Fix/test manifest view

buehlere:
This PR addressing a few outstanding issues with the manifest view and operator: 
- Updated `construct_csv` used by view and operator. [Query was not strict enough, causing extra bams to be included when generating the manifest](https://github.com/mskcc/beagle/pull/1233/files#diff-d7c68f21a9beedfc9cfd2e77d716676f00817d099e22306d525c7438b4ef357a). This was causing tests to fail as output from `construct_csv` was not reliable. 
- The [fixtures needed overhauled](https://github.com/mskcc/beagle/pull/1233/files#diff-ce246314b8d647341e6b045556b54338d450b5479415cd5bb07881ea6e0e1799) for the manifest view test.
- [Updates to the access manifest operator](https://github.com/mskcc/beagle/pull/1233/files#diff-da941693e8c9c8122c35a431065f0daf098c7fcdc8659ebc557cc3b499f564c7) after testing on dev. The major change here was to have the operator use a [pass through cwl](https://github.com/msk-access/cwl_pass_through/blob/main/manifest.cwl) to pass the output it generated to the `RunCreator`.
  - Also needed to update the test to accommodate the changes.


file_system/test_manifest_view.py
```diff
@@ -26,26 +26,25 @@ class TestManifest(TestCase):
         "28ca34e8-9d4c-4543-9fc7-981bf5f6a97f.files.json",
```
sivkovic, line 26: Can we move this test under `file_system/tests/test_manifest_view.py`, and also move `tests.py` (also rename it in something like tests_file_system.py) under the same directory. It is a good practice when having multiple test files to keep them in `tests` package in the app, like we did here https://github.com/mskcc/beagle/tree/master/runner/tests

sivkovic (changes requested):
Also not in this PR 
 but classnames should start with uppercase. `https://github.com/mskcc/beagle/blob/master/file_system/serializers.py#L536`.

Uppercase here as well `https://github.com/mskcc/beagle/blob/master/file_system/views/file_view.py#L287`
and move it to new `*view.py` file.

`https://github.com/mskcc/beagle/blob/master/runner/tasks.py#L894` this should be done through `FileRepository`.

Uppercase class name:
https://github.com/mskcc/beagle/blob/master/runner/tasks.py#L894
and move the logic into `file_system/helper/access_helper.py`

file_system/test_manifest_view.py
buehlere: I believe I've resolved these issues in the most recent commit. I double check that the changes work on dev as well. Let me know your thoughts!

sivkovic (approved):

PR #1239, opened 2023-06-13
Set for restart fix

sivkovic:


nikhil (changes requested):
It actually should be resume, because we are storing the last run info
so on the first restart then the last run was a resume

allanbolipata (approved):

PR #1343, opened 2024-07-03
Feature/add mapping data clinical endpoint

allanbolipata:
Adds three new endpoints that gives string outputs for mapping, pairing, and data clinical values based on data that's in beagle at runtime.


runner/views/run_api_view.py
```diff
@@ -652,6 +653,82 @@ def post(self, request):
...
+
+class ArgosDataClinicalViewSet(GenericAPIView):
+    logger = logging.getLogger(__name__)
+
+    serializer_class = ArgosPairingSerializer
+
+    def post(self, request):
+        igo_request_id = request.data.get("igo_request_id", None)
+        argos_slug = request.data.get("argos_slug", None)
+        if igo_request_id and argos_slug:
```
sivkovic, line 691: I think this is always True because in serializer both values can't be blank.

runner/views/run_api_view.py
```diff
@@ -652,6 +653,82 @@ def post(self, request):
...
+                if is_dmp_sample:
+                    dmp_samples.append(this_sample)
+            argos_inputs, error_samples = construct_argos_jobs(samples)
+            sample_mapping, filepaths = operator.get_mapping_from_argos_inputs(argos_inputs)
+            dmp_samples = operator.get_dmp_samples_from_argos_inputs(argos_inputs)
+            pipeline = operator.get_pipeline_id()
+            try:
+                pipeline_obj = Pipeline.objects.get(id=pipeline)
+            except Pipeline.DoesNotExist:
+                pass
```
sivkovic, line 714: return Response with error for example 404. Otherwise it will break https://github.com/mskcc/beagle/pull/1343/files#diff-ae1940d4baf50cb6d3ef8747783319716e5256c420bdd3f81be25bf51428f6dbR717 with 500

runner/views/run_api_view.py
```diff
@@ -652,6 +653,82 @@ def post(self, request):
...
+    serializer_class = ArgosPairingSerializer
+
+    def post(self, request):
+        igo_request_id = request.data.get("igo_request_id", None)
+        argos_slug = request.data.get("argos_slug", None)
+        if igo_request_id and argos_slug:
+            operator_model = Operator.objects.get(slug=argos_slug)
+            operator = OperatorFactory.get_by_model(operator_model, request_id=igo_request_id)
+            # construct_argos_jobs() is sloppily separate from the Operator module
+            from runner.operator.argos_operator.v2_0_0.construct_argos_pair import construct_argos_jobs
```
sivkovic, line 668: Is this import here on purpose? Should we move it on top of the file with other imports?

sivkovic (approved):

runner/views/run_api_view.py
allanbolipata: The problem with these is they're tied to the argos operator and that specific version

It's really messy, I'd prefer to not have it up top for now

runner/views/run_api_view.py
allanbolipata: Added

runner/views/run_api_view.py
allanbolipata: Removed

PR #1365, opened 2024-10-04
Feature/nucleovar operator

buehlere:
- Adding nextflow operator for nucleovar pipeline
- option to create input json from nf-core schemas
- passing nfcore template parameter to ridgeback


runner/run/objects/nextflow/nextflow_run_object.py
```diff
@@ -8,9 +8,11 @@
 from runner.run.objects.nextflow.nextflow_port_object import NextflowPortObject
 from runner.exceptions import PortProcessorException, RunCreateException, RunObjectConstructException
 
+logger = logging.getLogger(__name__)
+
 
 class NextflowRunObject(RunObject):
-    logger = logging.getLogger(__name__)
+    # logger = logging.getLogger(__name__)
```
sivkovic, line 15: We can probably remove this if it is not used anywhere

runner/pipeline/pipeline_cache.py
```diff
@@ -16,7 +16,7 @@ def get_pipeline(pipeline):
             resolved_dict = _pipeline.get("app")
         else:
             resolver_class = PipelineCache._get_pipeline_resolver(pipeline.pipeline_type)
-            resolver = resolver_class(pipeline.github, pipeline.entrypoint, pipeline.version)
+            resolver = resolver_class(pipeline.github, pipeline.entrypoint, pipeline.nfcore_template, pipeline.version)
```
sivkovic, line 19: Can you re-order params so nfcore_template is last one and put nfcore_template as param with default value, since I am not sure this will work for CWL pipelines.

sivkovic (approved):

PR #1403, opened 2025-01-22
Add new copy outputs operator to handle disambiguate output files

nikhil:


runner/operator/copy_outputs_operator/v2_1_0/copy_outputs_operator.py
```diff
@@ -0,0 +1,130 @@
...
+    def register_tmp_file(self, path):
+        fname = os.path.basename(path)
+        temp_file_group = FileGroup.objects.get(slug="temp")
+        file_type = FileType.objects.get(name="txt")
+        try:
+            File.objects.get(path=path)
+        except:
+            print("Registering temp file %s" % path)
+            f = File(file_name=fname, path=path, file_type=file_type, file_group=temp_file_group)
+            f.save()
```
sivkovic, line 111: This is a wrong way of creating the file. Without the FileMetadata these files will be unsearchable by FileRepostory. We should probably do something like this https://github.com/mskcc/beagle/blob/master/beagle_etl/jobs/metadb_jobs.py#L1101-L1104

nikhil: what metadata should go into that file, just project_prefix?

nikhil: okay added the changes in!

sivkovic (approved):

PR #1420, opened 2025-03-10
Hotfix/snp tumor search

buehlere:
A few patches for cci operators:
- change matched tumor and fillout search to account for `-T0` as well as `-L0` in the v1 and v2 snv operators
- add static bed to nucleo heme
- add chipvar operator for heme


sivkovic (approved):

allanbolipata:
Pausing merge on this until we get more guidance from customer

PR #1446, opened 2025-06-10
Feature/general import trigger

buehlere:
Starting an operator based on any combination of request level fastq metadata.

- add `recipes_json` to operator model, which specifies a list of dictionary each containing a valid combination of metadata for triggering an operator.
- add `fastq_metadata` to requestcallbackjob model
- match incoming `fastq_metadata`  from a cmo-request-update, cmo-sample-update and cmo-new-request to operator `recipes_json`, and triggering operator runs for matches. For example,
`fastq_metadata`: `{"genePanel": "val2", "baitSet": "baiSet2"}` would match the second dictionary in the recipes_json: `[{"genePanel": ["val1","val3"], "baitSet": "baiSet1"}, {"genePanel": "val2", "baitSet": "baiSet2"}]` for operator1.
- explicit when looking for imported fastq metadata (`file_group=settings.IMPORT_FILE_GROUP`), we don't want to consider other types of files.


scripts/remove_output_dirs.py
```diff
@@ -0,0 +1,53 @@
+import logging
```
sivkovic, line 1: This code should probably be part of some other PR?

sivkovic (approved):

PR #1528, opened 2026-06-23
Feature/force import

buehlere:
adds an API endpoint `request/force-import/<str:request_id>/` that allows a request to bypass metadata checks after it's SMILE message failed to import.

### DIFF
--- a/file_system/urls.py
+++ b/file_system/urls.py
@@ -33,5 +33,5 @@ urlpatterns = [
     path("batch-patch-files", BatchPatchFiles.as_view()),
     path("copy-files", CopyFilesView.as_view()),
     path("manifest/", Manifest.as_view()),
-    path("request/force-import/<str:request_id>/", ForceImportRequestView.as_view()),
+    path("request/force-import/<str:id>/", ForceImportRequestView.as_view()),
 ]
--- a/file_system/urls.py
+++ b/file_system/urls.py
@@ -11,7 +11,7 @@ from file_system.views.file_metadata_view import FileMetadataView
 from file_system.views.file_type_view import FileTypeView
 from file_system.views.sample_view import SampleViewSet
 from file_system.views.patient_view import PatientViewSet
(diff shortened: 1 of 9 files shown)


file_system/views/request_view.py
```diff
@@ -16,3 +21,24 @@ class RequestViewSet(
...
+            SMILEMessage.objects.filter(request_id=request_id, topic=settings.METADB_NATS_NEW_REQUEST)
+            .order_by("-created_date")
+            .first()
+        )
+        if not message:
+            return Response(
+                {"detail": f"No new-request SMILEMessage found for request_id {request_id}."},
+                status=status.HTTP_404_NOT_FOUND,
+            )
+        new_request(str(message.id), force_import=True)
```
sivkovic: It is probably better to run this as a task `new_request.delay(str(message.id), force_import=True)`

file_system/urls.py
```diff
@@ -33,4 +33,5 @@
     path("batch-patch-files", BatchPatchFiles.as_view()),
     path("copy-files", CopyFilesView.as_view()),
     path("manifest/", Manifest.as_view()),
+    path("request/force-import/<str:request_id>/", ForceImportRequestView.as_view()),
```
sivkovic: Can we change to url to point to `beagle_etl`. You can name it `import` or `etl`. We already have `request `endpoints but for different purpose https://github.com/mskcc/beagle/blob/master/file_system/urls.py#L26.

sivkovic (changes requested):
We should also have this functionality through admin panel. You can add it as action in `beagle_etl.admin.SMILEMessagesAdmin`

buehlere:
> We should also have this functionality through admin panel. You can add it as action in `beagle_etl.admin.SMILEMessagesAdmin`

I added these changes. Let me know what you think. It's live on staging if you want to try it out: http://isvvoyagerstage01.mskcc.org:4007/

sivkovic (approved):

sivkovic (approved):

PR #1537, opened 2026-07-13
Import retry

sivkovic:
## Changelog

- Implemented retry logic when processing SMILE Import
- Send email notification in case import fails because of the wrong fastq permissions


beagle_etl/jobs/metadb_jobs.py
```diff
@@ -214,6 +215,26 @@ def new_request(message_id):
     study, _ = Study.objects.get_or_create(study_id=StudyObject.generate_study_id(data.labHeadName))
 
     valid_samples = {k for k, v in status.items() if v.status == "COMPLETED"}
+
+    retry_samples = {k for k, v in status.items() if v.status == "RETRY"}
+
+    if retry_samples:
+        sample_status = sorted([sample.to_dict() for sample in status.values()], key=lambda d: d["sample"])
+        message.set_sample_status(sample_status)
+        message.add_log(f"Permission Denied error during import for igoRequestId:{message.request_id} id:{message_id}")
```
allanbolipata, line 232: Can we break this down further for logging and standardize - like why the import failed, which sample(s) failed, the reason, the number of files, etc. This will make it easier to grep or find in logs.

Something like

IMPORT_PERMISSION_FAIL: Sample `sample_id` failed import. Skipped {num_files} out of {num_files_total} due to {whatever}.

allanbolipata (approved):

